PR #16, opened 2023-09-19
Ds/update method

dennispelle:
Config and Lang will be downlaoded if not found.
no scrollbars for buttonstripe
Buttonsize can be typed in and buttonstripe will redadjust
deleted some not needed buttons
dropdown for measuretype in save
upload successfull popup after data send


NP199:
new Pull request for merge

PR #40, opened 2023-10-20
Adds Workflow File to add a release if code is pushed with a tag

bjoekeldude:
In order to have the newest Version displayed on the front page of this project, I wanted to have the CI create a release. 

If one pushes a tag, now a new release gets generated.


dominicpoeschko (approved):
looks good to me. Maybe it would be possible to merge the workflow files so the binaries get only build ones.

PR #49, opened 2023-10-25
added cmake version info and git hash.

dominicpoeschko:
The CMake version and git hash can now be seen in the GUI.


bjoekeldude (commented):
I have another idea for improvement, but this should already be merged.

bjoekeldude (approved):

PR #65, opened 2023-11-13
Defined a callback to filter entered characters for VIN

R-Abbasi:
- Defined a callback to filter entered data for VIN, based on a list of 17 legal characters specified as regexes.


bjoekeldude (changes requested):
please use the `fifth_to_seventh` policy for all characters.

bjoekeldude (approved):

PR #77, opened 2023-12-21
Added desgin and functionality to save devices' data

R-Abbasi:
Modified `main.cpp`, `look_up_saves.hpp`, `imfilebrowser.h` and `saves_popup.hpp` files in addition to increasing code readability, as follows respectively: 
- Greyed out the _Speicher_ tool-button when the  _Speichern der aufgenommenen Daten_ pop-up is open and got it return to normal style when it's closed
- Modified `getSubdirectoriesInFolder(...)` to take `char*`s to deal with and show _scantype_ and _mileage_ input texts
- Added functionality to select a desired path using the OS explorer to be used to save devices data into, instead of _StoragePathInputField1_'s path
- Rearranged `InputText`s and (added) buttons with _Tooltip_
- Added functionality to save devices data into various paths and set current directory to `../Omniview`
- Changed `imfilebrowser.h`: https://github.com/skunkforce/OmniView/issues/76

Issue #75, opened 2023-12-18
Programm crash after second measurement 


AKMaily:
When clicking on 'Search for devices' after the initial measurement has been reset, the window closes, and the program crashes.

NP199:
I cant reproduce this error when pressing "Search for devices". 
I have an error when saving the measurement, the window closes and the programm crashes.

AKMaily:
1.  Did you record a measurement, reset it and then try to record again? This has already crashed on two PCs.
  2.  Did you get an error message when the programm crashes ?

________________________________
Von: Niklas Pierick ***@***.***>
Gesendet: Mittwoch, 20. Dezember 2023 17:39
An: skunkforce/OmniView ***@***.***>
Cc: AKMaily ***@***.***>; Author ***@***.***>
Betreff: Re: [skunkforce/OmniView] Programm crash after second measurement (Issue #75)


I cant reproduce this error when pressing "Search for devices".
I have an error when saving the measurement, the window closes and the programm crashes.

NP199:
Yes i tried reproducing the error, as you wrote it down. I resetted the thing mutliple times and used both functions. I had only problems when using the save function.
I used the githash version [ec38fa6]



My colleague used the githash version [0836b9b] (https://github.com/skunkforce/OmniView/commit/0836b9b19051e04c5ca148a99dc150090d91e5a7) yesterday and had no problems with searching and saving the files.

I got no error messages

AKMaily:
Did you use the windows or the linux version?

NP199:
My colleague and myself both used the windows version

AKMaily:
Did you connect one or two devices to the pc?

AKMaily:
This issue has been fixed with Pull Request #91.

PR #87, opened 2024-01-22
features designing and data savings

R-Abbasi:
- Added lots of changes/features for design and functionality.
For more details, take a look at the description and commits of #77 
And more ...


AKMaily (approved):
Still Issues with connecting two devices, so the saving process isnt completly tested yet (Bug #75). The rest works well.

PR #91, opened 2024-01-29
improvement regarding handling multi-devices

R-Abbasi:
- Improved the code to handle saving data for more than one connected devices


AKMaily (approved):
Fixes the issue.

Issue #92, opened 2024-01-29
Refactoring files in OmniView/src


R-Abbasi:
A suggestion to refactor files in _OmniView/src_, enhance the readability and make future maintenance more straightforward/easier:
- create two libraries (more preferably using `#ifndef ...` and not `#pragma once` include guards) managing related parts of the whole functionality this way: `popups.hpp` as the interface for: `settingspopup`, `saves_popup`, `look_up_saves` and `create_training_data`, and `handler.hpp` as the interface for: `apihandler`, `get_from_gthub`, `jasonhandler` and `sendData`. That is, two interfaces with 8 implementations (`.cpp` files) and `include` those two libraries in `main.cpp` instead of 8 header-only libraries. This way, we can get rid of header-only libraries and their disadvantages (look it up or [read this, for example](https://stackoverflow.com/a/12673877/4989541)).

- It is also possible to simplify `main.cpp` by taking part of its code into `handler.hpp` and declare the needed function(s) there.

NP199:
Great suggestion.
I don`t understand the reason to use #ifndef ... over pragma once. Can you elaborate on this?

What do you think about the use of designated namespaces in the libraries?

R-Abbasi:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf8-use-include-guards-for-all-header-files:
_Some implementations offer vendor extensions like #pragma once as alternative to include guards. It is not standard and it is not portable. It injects the hosting machine’s filesystem semantics into your program, in addition to locking you down to a vendor._ So, I assume using `#ifndef ...` might be a better choice.

Namespaces as a mechanism for expressing that some declarations belong together and that their names shouldn’t clash with other names sound harmless but considering the libraries I've worked on (_OmniView/src_, `imfilebrowser.h` and part of _/OmniView/ai_omniscope-v2-communication_sw/src_), using namespaces looks to be for that reason or an approach to make the provided features seem as if they're built-in `ImGui` ones, e.g., `ImGui::FileBrowser` in `imfilebrowser.h`. In `main.cpp`, however, using that namespace doesn't make much sense to me. 
Anything you like to add/modify, please?

R-Abbasi:
[...]

PR #94, opened 2024-02-03
using separate interfacse and implementation, design and functionality changes

R-Abbasi:
- Used the approach suggested in Issue #92 to separate interfaces (`.hpp` files) from their implementations
- Added an interface having a number of functions and implemented them as new `.cpp` files
- Modified some of the headers required for the functionality of _generate training data_ menu 

I see room for further improvements and this `pr` is to address part of the given tasks and a sample for the suggested changes in the project and its codebase

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,12 @@ endif()
 include(cmake_helpers/BuildOptions.cmake)
 include(cmake_helpers/FindOrFetch.cmake)
 
-add_executable(OmniView src/main.cpp)
+add_executable(OmniView 
+               src/main.cpp 
+               src/generateTrainingData.cpp 
+               src/save_warning_popup.cpp
+               )
+
 
 target_add_default_build_options(OmniView PUBLIC)
 
--- a/src/generateTrainingData.cpp
+++ b/src/generateTrainingData.cpp
@@ -1,0 +1,155 @@
+#include "popups.hpp"
+#include "sendData.hpp"
+#include <iostream>
+#include <regex>
+#include <sstream>
+
+void generateTrainingData(
+    bool &open_generate_training_data,
+    const std::vector<std::shared_ptr<OmniscopeDevice>> &devices,
+    const std::vector<std::string> &savedFileNames) {
+
+  const size_t devicesSz{devices.size()};
+
+  //  one extra space for '\0' character and another
+  //  for one past the last accepted input character
+  static char vinBuffer[19];
+  auto vinFilter = [](ImGuiInputTextCallbackData *data) -> int {
+    const std::regex chars_regex("[A-HJ-NPR-Z0-9]");
+    std::string s;
+    // get entered char and save it into string
+    s += data->EventChar;
+    // strlen is updated when entered char passes the filter
+    size_t indx = strlen(vinBuffer) + 1;
+
+    if (indx >= 1 && indx <= 17)
+      return !std::regex_match(
+          s, chars_regex); // return 0 as passed for matched chars
+    return 1;              // discard exceeding chars
+  };
+
+  ImGui::OpenPopup("Generate Training Data");
+
+  if (ImGui::BeginPopupModal("Generate Training Data", nullptr,
+                             ImGuiWindowFlags_AlwaysAutoResize)) {
+
+    // Have address of bool for std::vector
+    struct BoolWrapper {
+      BoolWrapper() : b(false) {}
+      BoolWrapper(bool _b) : b(_b) {}
+      bool b;
+    };
+
+    // devices checkboxes buffer
+    static std::vector<BoolWrapper> dvcChackedArr(devicesSz, false);
+    dvcChackedArr.resize(devicesSz);
+
+    // file names checkboxes buffer
+    static std::vector<BoolWrapper> flnmChackedArr(devicesSz, false);
+    flnmChackedArr.resize(devicesSz);
+
+    std::stringstream ss;
+
+    static bool ucw = false;
+    static bool wUcwPop = false;
+    if (ImGui::RadioButton("User current Waveform", ucw))
+      ucw = !ucw;
+    if (ucw && !devices.size())
+      wUcwPop = true;
+    else if (ucw && devices.size()) {
+      if (ImGui::BeginCombo("##ComboDevice", "Devices Menu")) {
+        for (size_t i = 0; i < devicesSz; i++) {
+          ss << devices[i]->getId().value().serial;
+          ImGui::Checkbox(ss.str().c_str(), &((dvcChackedArr.begin() + i)->b));
+          ss.str(std::string());
+        }
+        ImGui::EndCombo();
+      }
+    }
+    if (wUcwPop) {
+      save_warning_popup(wUcwPop, "No Waveforms were made");
+      ucw = false;
+    }
+
+    static bool wff = false;
+    static bool wWffPop = false;
+    if (ImGui::RadioButton("Waveform from File", wff))
+      wff = !wff;
+    if (wff && !savedFileNames.size())
+      wWffPop = true;
+    else if (wff && savedFileNames.size()) {
+      if (ImGui::BeginCombo("##ComboFile", "Waveforms Menu")) {
+        for (size_t i = 0; i < savedFileNames.size(); i++)
+          ImGui::Checkbox(savedFileNames[i].c_str(),
+                          &((flnmChackedArr.begin() + i)->b));
+        ImGui::EndCombo();
+      }
+    }
+    if (wWffPop) {
+      save_warning_popup(wWffPop, "No files were saved during measurement");
+      wff = false;
+    }
+
+    static char ID[10];
+    static char milage[10];
+    ImGui::SetNextItemWidth(300); // custom width
+    ImGui::InputTextWithHint("ID", "Enter ID(optional)", ID, IM_ARRAYSIZE(ID));
+    ImGui::SetNextItemWidth(300);
+    ImGui::InputTextWithHint(
+        "VIN", "Enter VIN", vinBuffer, IM_ARRAYSIZE(vinBuffer),
+        ImGuiInputTextFlags_CharsUppercase | ImGuiInputTextFlags_CharsNoBlank |
+            ImGuiInputTextFlags_CallbackCharFilter,
+        // callback function to filter each character
+        // before putting it into the buffer
+        vinFilter);
+    ImGui::SetNextItemWidth(300);
+    ImGui::InputTextWithHint("milage", "Enter milage", milage,
+                             IM_ARRAYSIZE(milage));
+
+    std::string msg{ID};
+    // have each entry on a new line
+    msg += '\n';
+    msg += vinBuffer;
+    msg += '\n';
+    msg += milage;
+
+    ImGui::SeparatorText("Reason-for-investigation");
+    static int b = 0;
+    ImGui::RadioButton("Maintenance", &b, 0);
+    ImGui::SameLine();
+    ImGui::RadioButton("Fault", &b, 1);
+
+    ImGui::SeparatorText("Electrical-Consumers");
+    static int c = 0;
+    ImGui::RadioButton("Off", &c, 0);
+    ImGui::SameLine();
+    ImGui::RadioButton("On", &c, 1);
+
+    ImGui::SeparatorText("Assessment");
+    static int d = 0;
+    ImGui::RadioButton("Normal", &d, 0);
+    ImGui::SameLine();
+    ImGui::RadioButton("Anomaly", &d, 1);
+
+    static char comment[16];
+    ImGui::InputTextMultiline("Comment", comment, IM_ARRAYSIZE(comment),
+                              ImVec2(250, 70),
+                              ImGuiInputTextFlags_AllowTabInput);
+
+    if (ImGui::Button("Cancel")) {
+      ImGui::CloseCurrentPopup();
+      open_generate_training_data = false;
+    }
+
+    ImGui::SameLine();
+    if (ImGui::Button(" Send ")) {
+      // example url
+      const std::string url{
+          "https://raw.githubusercontent.com/skunkforce/omniview/"};
+      sendData(msg, url);
+      ImGui::CloseCurrentPopup();
+      open_generate_training_data = false;
+    }
+    ImGui::EndPopup();
+  }
+}
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -14,8 +14,8 @@
 #include <fmt/chrono.h>
 #include <fmt/core.h>
 #include <fmt/format.h>
-#include <fstream>
-#include <imgui.h>
+//#include <fstream>
+//#include <imgui.h>
 #include <nlohmann/json.hpp>
 #include <nlohmann/json_fwd.hpp>
 #include <set>
@@ -24,10 +24,8 @@
 // clang-format off
 #include <imfilebrowser.h>
 // clang-format on
-#include "sendData.hpp"
-#include <regex>
+#include "popups.hpp"
 
-namespace ImGui {
 inline void SetupImGuiStyle(bool bStyleDark_, float alpha_) {
   ImGuiStyle &style = ImGui::GetStyle();
 
@@ -113,7 +111,6 @@ inline void SetupImGuiStyle(bool bStyleDark_, float alpha_) {
     }
   }
 }
-} // namespace ImGui
 
 static std::vector<std::string>
 getAvailableLanguages(std::string const &languageFolder) {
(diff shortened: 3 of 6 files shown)


CMakeLists.txt
```diff
@@ -29,7 +29,12 @@ endif()
 include(cmake_helpers/BuildOptions.cmake)
 include(cmake_helpers/FindOrFetch.cmake)
 
-add_executable(OmniView src/main.cpp)
+add_executable(OmniView
```
NP199, line 32: I would suggest to seperate the libraries and include them to the build with "add_library"

src/generateTrainingData.cpp
```diff
@@ -0,0 +1,155 @@
...
+    if (ImGui::Button("Cancel")) {
+      ImGui::CloseCurrentPopup();
+      open_generate_training_data = false;
+    }
+
+    ImGui::SameLine();
+    if (ImGui::Button(" Send ")) {
+      // example url
+      const std::string url{
+          "https://raw.githubusercontent.com/skunkforce/omniview/"};
```
NP199, line 148: Shouldn`t there be a real url to upload the data?

src/save_warning_popup.cpp
```diff
@@ -0,0 +1,20 @@
+#include "popups.hpp"
+#include <iostream>
+
+void save_warning_popup(bool &popup, const std::string &msg) {
+
+  ImGui::OpenPopup("Save warning");
```
NP199, line 6: This a generic warning and save warning suggests that you failed to save something or something similiar.
At least i get the text "Save Warning" displayed in the head of the popup if i press the radio button "User current Waveform".
```suggestion
  ImGui::OpenPopup("Warning");
```

src/saves_popup.hpp
```diff
@@ -1,336 +1,302 @@
...
+  };
+
+  // connected devices at runt-time
+  const size_t devicesSz{devices.size()};
+  static constexpr size_t inptTxtArrSz{100};
+
+  static std::vector<std::string> savedFileNames;
+
+  // buffer array for the input text field(s)
+  static std::vector<std::vector<char>> inptTxtFilds(
```
NP199, line 88:
```suggestion
  static std::vector<std::vector<char>> inptTxtFields(

```

src/saves_popup.hpp
```diff
@@ -1,336 +1,302 @@
...
+    for (size_t i = 0; i < devicesSz; i++) {
+      ss << "Device " << i + 1;
+      ImGui::Checkbox(ss.str().c_str(), &((dvcChackedArr.begin() + i)->b));
+      ss.str(std::string());
+    }
+    ImGui::EndCombo();
+  }
+
+  static char scantype[255] = "";
+  static char vin[18] = "";
```
NP199, line 132: is it intentional that vin only has 18 character?

Previous you used 19 in "generate_trainig_data":"
  //  one extra space for '\0' character and another
  //  for one past the last accepted input character
  static char vinBuffer[19];"

src/generateTrainingData.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    // Have address of bool for std::vector
+    struct BoolWrapper {
+      BoolWrapper() : b(false) {}
+      BoolWrapper(bool _b) : b(_b) {}
+      bool b;
+    };
+
+    // devices checkboxes buffer
+    static std::vector<BoolWrapper> dvcChackedArr(devicesSz, false);
```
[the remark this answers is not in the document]
R-Abbasi: Yes, `Checked` not `Chacked`.

src/generateTrainingData.cpp
```diff
@@ -0,0 +1,155 @@
...
+    dvcChackedArr.resize(devicesSz);
+
+    // file names checkboxes buffer
+    static std::vector<BoolWrapper> flnmChackedArr(devicesSz, false);
+    flnmChackedArr.resize(devicesSz);
```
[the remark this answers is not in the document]
R-Abbasi: It's stated in the comment right above that.

src/generateTrainingData.cpp
```diff
@@ -0,0 +1,155 @@
...
+        }
+        ImGui::EndCombo();
+      }
+    }
+    if (wUcwPop) {
+      save_warning_popup(wUcwPop, "No Waveforms were made");
+      ucw = false;
+    }
+
+    static bool wff = false;
```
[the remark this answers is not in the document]
R-Abbasi: A comment was added:
```
 // Waveform From File
 static bool wff = false;
```

src/generateTrainingData.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    std::string msg{ID};
+    // have each entry on a new line
+    msg += '\n';
+    msg += vinBuffer;
+    msg += '\n';
+    msg += milage;
+
+    ImGui::SeparatorText("Reason-for-investigation");
+    static int b = 0;
```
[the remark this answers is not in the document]
R-Abbasi: This part dates back to about two months ago not touched in the current task. Yet I agree that a single character might not be always a good suggestion in a declaration although it seemingly is not that uncommon in the Docs e.g., the first example for `ImGui::RadioButton` in the demo or `e` and `g` [here] and more. (https://github.com/ocornut/imgui/blob/docking/imgui.cpp#L1383).

`b`, as an example, is used in the code only for two statements right below the declaration. Still changing it to e.g., `invest` may look nicer. 
As for using a `bool`, I toyed with that a little but didn't find a simpler version for those three pairs that way.

CMakeLists.txt
R-Abbasi: Yes, and thought of it before sending the PR but guessed it'd be good when the number of `.cpp` files will grow later on.

src/save_warning_popup.cpp
R-Abbasi: Agreed.

src/saves_popup.hpp
R-Abbasi: The functionality was correct to get only 17 characters - tested for its task at the time - but it seems to have one innocent extra character space.

src/look_up_saves.hpp
```diff
@@ -10,15 +10,15 @@
...
+                                      char inputvin[18] = 0,
+                                      char mileage[10] = 0) {
   bool deviceFields = false;
   if (scantype != 0 && inputvin != 0 && mileage != 0)
     deviceFields = true;
 
   std::vector<std::string> subdirectories;
-  subdirectories.push_back(
-      load_json<std::string>(language, "savepopup", "new_car"));
+  subdirectories.push_back("New Car");
```
AKMaily, line 21: Why did you change this part?

R-Abbasi: Stephan suggested using English terms and that was a case I saw when dealing with the task.

AKMaily: Using English terms is right. But in a later process the language is loaded from the language.json file and the user can choose between different languages, so i think it would be better to keep the load_json parts.

R-Abbasi: Probably it can be possible to use a `std::map` or one of its flavors instead. That way, it sounds even to be more performant than loading a file multiple times, I guess.

src/generateTrainingData.cpp
```diff
@@ -1,62 +1,81 @@
...
       ucwWPop = true;
-    else if (ucw && devicesSz) {
-      if (devicesSz != savedFileNames.size())
-        fmt::println("size of devices and waveforms don't match!");
-      else if (ImGui::BeginCombo("##ComboDevice", "Devices & Waveforms Menu")) {
-        for (size_t i = 0; i < devicesSz; i++) {
-          ImGui::Checkbox(devices[i]->getId().value().serial.c_str(),
-                          &(dvcCheckedArr[i].b));
-          ImGui::Checkbox(savedFileNames[i].c_str(), &(flnmCheckedArr[i].b));
+    else if (ucw && (devicesSz || savedFileNames.size()) &&
```
AKMaily, line 39: There should be no variable called 'b' or 'k' if they are not used as counters in loops.

R-Abbasi: Not sure I got your statement well. Could you explain it a little more, please. 
Both `b` and `k` are used in the loops and that part of code is to address the section: _The User should only be able to select ONE waveform in the “Use current waveform” drop down menu_ of the current tasks.

src/main.cpp
```diff
@@ -213,19 +207,15 @@ int main() {
   double xmax_paused{0};
   static bool open_settings = false;
   static bool open_generate_training_data = false;
-  static bool open_save_devices = false;
   // unique and ordered filenames
-  static std::set<std::string> savedFileNames;
+   std::set<std::string> savedFileNames;
```
AKMaily, line 211: why is this not static anymore ?

src/main.cpp
```diff
@@ -213,19 +207,15 @@ int main() {
...
-  static std::set<std::string> savedFileNames;
+   std::set<std::string> savedFileNames;
   static bool upload_success = false;
   static bool flagPaused = true;
   static bool flagDataNotSaved = true;
   static ImVec2 mainMenuBarSize;
   std::optional<OmniscopeSampler> sampler{};
   std::map<Omniscope::Id, std::vector<std::pair<double, double>>> captureData;
 
-  std::string path;
```
AKMaily: Why did you delete this ?

src/popups.hpp
```diff
@@ -18,11 +20,61 @@
...
+       {"AnalyzeData", "Daten\nanalysieren"},
+       {"Crt_trng_data", "Erstellen\nSie\nTrainingsdaten"},
+       {"Back", "Zurück"} };
+
+static auto appLanguage = englishLan;  
+static ImGui::FileBrowser fileBrowser;
+static ImGui::FileBrowser directoryBrowser(ImGuiFileBrowserFlags_SelectDirectory);    
 void generateTrainingData(bool&, 
-                    const std::vector<std::shared_ptr<OmniscopeDevice>>&,
+                    const std::optional<OmniscopeSampler>&,
```
AKMaily, line 76: Why did you use sampler instead of a device ?

src/saves_popup.hpp
```diff
@@ -240,57 +196,57 @@ auto saves_popup(nlohmann::json const &config, nlohmann::json const &language,
...
       ImGui::PopStyleVar();
       ImGui::SetItemTooltip("Add another path");
-
-      ImGui::Dummy({750, 0});
-      ImGui::SameLine();
     }
 
     // to save captureData from main.cpp into file
-    if (ImGui::Button("save", ImVec2(load_json<Size>(config, "button")))) {
+    if (ImGui::Button(" save ")) {
```
AKMaily, line 213: Language file ?

src/saves_popup.hpp
```diff
@@ -240,57 +196,57 @@ auto saves_popup(nlohmann::json const &config, nlohmann::json const &language,
...
+        ++i;
       }
-      indxArr.clear();
       ImGui::CloseCurrentPopup();
     }
     ImGui::SameLine();
   }
 
-  if (ImGui::Button("back", {0, 0}))
+  if (ImGui::Button(" back "))
```
AKMaily, line 250: Language File ?

src/generateTrainingData.cpp
```diff
@@ -3,84 +3,72 @@
...
-          if (b) {
-            k = i;
-            selectedItem = devices[i]->getId().value().serial.c_str();
-          } else {
-            k = -1;
-            selectedItem = "Devices & Waveforms Menu";
+      bool b{false};
+      if (sampler.has_value())
+        for (const auto &device : sampler->sampleDevices) {
+          b = (k == i);
```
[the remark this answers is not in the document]
R-Abbasi: The algorithm assures us only one checkbox out of two sets of device IDs/waveforms is selected at any time. `b` is to check if one of the checkboxes is checked and `k` is a helper variable. I think their names aren't very important and understanding the algorithm is more valuable.

src/generateTrainingData.cpp
```diff
@@ -3,84 +3,72 @@
...
-        if (ImGui::Checkbox(it->c_str(), &b)) {
-          if (b) {
-            k = i + sz;
-            selectedItem = it->c_str();
-          } else {
-            k = -1;
-            selectedItem = "Devices & Waveforms Menu";
+      if (const size_t sz = savedFileNames.size())
+        for (const auto &file : savedFileNames) {
+          b = (k == i + sz);
```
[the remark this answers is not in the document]
R-Abbasi: `sz` gets the size of `savedFileNames` and then control enters the range-based `for`-loop. `b` is evaluated by the condition `k == i + sz`.

src/generateTrainingData.cpp
```diff
@@ -3,84 +3,72 @@
...
-        ImGui::PushID(i + sz);
-        bool b = (k == i + sz);
-        if (ImGui::Checkbox(it->c_str(), &b)) {
-          if (b) {
-            k = i + sz;
-            selectedItem = it->c_str();
-          } else {
-            k = -1;
-            selectedItem = "Devices & Waveforms Menu";
+      if (const size_t sz = savedFileNames.size())
```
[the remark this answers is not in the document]
R-Abbasi: No.
A reference to `savedFileNames.size()` will be sent to `sz` making it equal to the size, then `sz` is checked for validity leading to `true` if it's none-zero and `false` otherwise.

src/generateTrainingData.cpp
```diff
@@ -3,84 +3,72 @@
...
-            selectedItem = "Devices & Waveforms Menu";
+      bool b{false};
+      if (sampler.has_value())
+        for (const auto &device : sampler->sampleDevices) {
+          b = (k == i);
+          if (ImGui::Checkbox(device.first->getId().value().serial.c_str(),
+                              &b)) {
+            if (b) {
+              k = i;
+              selectedItem = device.first->getId().value().serial.c_str();
```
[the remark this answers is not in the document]
R-Abbasi: Device ID.

src/main.cpp
R-Abbasi: The `std::set` is not part of the render function therefore it's created merely once, at program startup.

src/main.cpp
R-Abbasi: Their values in the language maps are used instead.

src/popups.hpp
R-Abbasi: The functionality is similar; devices in the sampler are used. 
The sampler looks like the handler of the devices. It initially `std::move`s the connected devices and manages them for a couple of works. I thought making use of the devices in it to be more reasonable than separating them from it.

src/saves_popup.hpp
R-Abbasi: No, it's just a string literal. I try applying the language maps for all parts of the application if that's useful to you too.

src/generateTrainingData.cpp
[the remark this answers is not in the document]
AKMaily: For better understanding of the algorithm, it is advisable to name variables based on their functionality. Additionally, include comments within the code to describe its purpose and logic. This is particularly important, among other reasons, because this project is an open-source project.

src/generateTrainingData.cpp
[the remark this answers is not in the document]
AKMaily: Please name sz --> size for better understandment. b, i also should have correct names

src/generateTrainingData.cpp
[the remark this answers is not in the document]
AKMaily: There should be a comment for that.

src/saves_popup.hpp
AKMaily: The language map should be consistently applied to all relevant sections.

src/generateTrainingData.cpp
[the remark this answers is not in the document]
R-Abbasi: `b` and `k` there work pretty much like `i` and `j` in `for`-loops. Hopefully `is_checked` and `token` rather than them respectively helps.

src/generateTrainingData.cpp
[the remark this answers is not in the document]
R-Abbasi: Like what? We already have `size_t` and `size()` there. Taking a glance at `const size_t sz = savedFileNames.size()` it easily reveals what `sz` is.

Do you consider renaming `std` or `cout` in `std::cout << ... ;` required?

src/saves_popup.hpp
R-Abbasi: Already created a Discussion about that and it's still in progress. Will try using that throughout the whole code as soon we reach a conclusion there or you want it.

src/generateTrainingData.cpp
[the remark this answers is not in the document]
R-Abbasi: `getId()` and `serial.c_str()` are pretty much expressive, that's why I didn't feel need for a comment. But sure.

AKMaily (approved):

PR #114, opened 2024-03-14
upload and json file format

R-Abbasi:
- coded the `.json` part to meet the required format
- used a `nlohmann::ordered_json` object rather than `nlohmann::json` to preserve the order
- code modifications regarding better functionality and performant
- used the new `url` to send data to

### DIFF
--- a/src/generateTrainingData.cpp
+++ b/src/generateTrainingData.cpp
@@ -101,9 +101,7 @@ void generateTrainingData(
     // one extra space for '\0' character
     static char VIN[18];
     static std::string Mileage;
-
-    std::vector<double> numbers; // measurement values
-    static nlohmann::json myJson;
+    static std::vector<double> file_measuring_vals; // measurement values
     static bool grayFields = false;
     // custom flags
     static auto measuGrayFlag = ImGuiInputTextFlags_None;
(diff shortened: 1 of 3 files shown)


src/generateTrainingData.cpp
```diff
@@ -359,34 +356,38 @@ void generateTrainingData(
...
-
-        myJson["Comment"] = comment;
+           assessmnt = "Normal";
+
+         std::vector<double> y_values; 
+         if(usr_curnt_wave)
+           y_values = std::move(crnt_measuring_vals);
+         else 
+          y_values = std::move(file_measuring_vals);
+         myJson["meta"] = { Measurement, VIN, Mileage, invest_reason,
```
NP199, line 385: I would suggest to pack the Meta Data into key value pairs and not in an array of just strings.

src/sendData.hpp
```diff
@@ -4,53 +4,47 @@
...
-
-    // Clean up
-    curl_easy_cleanup(curl);
-    curl_slist_free_all(header_list);
+
+  // init the winsock stuff (Windows)
+  curl_global_init(CURL_GLOBAL_ALL);
+
+  // get a curl handle
+  CURL* curl = curl_easy_init();
```
NP199, line 16: maybe we should start packing the pointer into std::unique_pointer or shared pointer or use a wrapper for the curl library to prevent memory leaks?

NP199 (approved):

src/generateTrainingData.cpp
R-Abbasi: Tested `std::vector<std::pair<std::string,std::string>>` before `push`ing changes, but seemingly the external function on the server (the API), doesn't accept that. Not the one below:
```
 myJson["meta"] = { {"Measurement", Measurement}, {"VIN", VIN}, {"Mileage", Mileage},
                    {"Investigation reason", invest_reason},
                    {"Electrical consumers", elec_consumer},
                    {"Assessment", assessmnt}, {"Comment", comment}
 };
```
So had to use that. Probably it's another thing which needs change in the API specification.

PR #116, opened 2024-03-21
receiving analyzed waveform and saving it

R-Abbasi:
- added a callback function to receive the response from the server 
- saved the response into a `.csv` file with a specific name inside a specific folder 
- few more minor improvements


AKMaily:
Is this still an open PR regarding #142 ?

R-Abbasi:
Not anymore as the updated version (#142) is now created.

PR #130, opened 2024-04-05
At/fix styles

AKMaily:
New PR because the old one had a git bug. 

Changes: Style.hpp, Style.cpp, main.cpp

Changed the style of the Menu with ImGui Colors and ImPlot Colors. 

Created a developer menu which can be set to visible via the "Development" boolean in the main function. 
Now you can set the colors for the different elements in this menu and can see how it affects the GUI.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -9,7 +9,7 @@
 #include <fmt/core.h>
 
 int main() {
-  bool Development = true; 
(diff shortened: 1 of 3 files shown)


AKMaily:
This Pull-Request will be changed to be directly merged into the master.

PR #131, opened 2024-04-06
Ak/refactoring style

AKMaily:
Set the Style of the refactored code to the old style we implemented. 

Now there is a development menu that can be set to visible via the "Development" variable. In this menu the developer can set the colors of the GUI directly and see how it affects the GUI.


src/main.cpp
```diff
@@ -1,15 +1,17 @@
...
 #include "popups.hpp"
+#include "settingspopup.hpp"
+#include "style.hpp"
+#include <boost/asio.hpp>
+#include <cmake_git_version/version.hpp>
+#include <fmt/chrono.h>
+#include <fmt/core.h>
 
 int main() {
+  bool Development = false;
```
R-Abbasi, line 13: Where is this variable going to be `true`, please?

src/style.cpp
```diff
@@ -8,71 +8,105 @@
...
-  style.Colors[ImGuiCol_PlotLinesHovered] = {1.f, 0.43f, 0.35f, 1.f};
-  style.Colors[ImGuiCol_PlotHistogram] = {0.9f, 0.7f, 0.f, 1.f};
-  style.Colors[ImGuiCol_PlotHistogramHovered] = {1.f, 0.6f, 0.f, 1.f};
-  style.Colors[ImGuiCol_TextSelectedBg] = {0.26f, 0.59f, 0.98f, 0.35f};
-  // style.Colors[ImGuiCol_ModalWindowDarkening] = {0.2f, 0.2f, 0.2f, 0.35f};
-
-  if (bStyleDark_)
+  style.FrameRounding = 5.0f;
+  style.Colors[ImGuiCol_Text] =
+      ImVec4(255 / 255.0f, 255 / 255.0f, 255 / 255.0f, 100 / 100.0f);
```
R-Abbasi, line 25: As already told you, this style doesn't look to be very appropriate.
It seems less readable (since it requires more space to understand the code) and less performant (since it performs a number of calculations (division) per frame in addition to a possible implicit conversion between `int` and `float`).

src/style.cpp
```diff
@@ -319,3 +391,183 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
...
+    // Add more color pickers for other style elements
+
+    ImGui::TreePop();
+  }
+
+  ImGui::End();
+}
+
+void UpdateColors(const std::vector<StyleElement> &styleElements,
+                  const ImVec4 *colors) {
```
R-Abbasi, line 547: In general, try avoid using raw pointers, as much as possible.

src/LoadImages.hpp
```diff
@@ -0,0 +1,86 @@
+// Fürs laden der Bilder
```
R-Abbasi: Why do you need another header, please? 
This also lacks `include` guard and correct code. It's a header only library as well.

images/DiagnosticIcon_data.h
```diff
@@ -0,0 +1,84 @@
+unsigned char DiagnosticIcon_png[] = {
```
R-Abbasi, line 1: Why did you use a header for each image again, please?

src/main.cpp
AKMaily: If you set it to true and want to use the Development menu, this is only possible for people that work on the code

images/DiagnosticIcon_data.h
AKMaily: I did not, i put them all in a Icon header file. Those are only there because it is easier to see which header belongs to which image and to also have the images in the project for development

src/main.cpp
R-Abbasi: Using `ImGui::ColorEdit` and that color picker seems nice but much code has been added for that and I can't be sure why we should need changing the colors often.

src/style.cpp
```diff
@@ -97,6 +132,43 @@ void set_button_style_to(const nlohmann::json &config,
...
+  // style.Colors[ImGuiCol_ButtonHovered] = { 0.94f, 0.94f, 0.94f, 1.0f };
+  // style.Colors[ImGuiCol_ButtonActive] = { 0.94f, 0.94f, 0.94f, 1.0f };
+
+  // plotStyle.Colors[ImPlotCol_PlotBg]= { 1.0f, 1.0f, 0.93f, 1.0f };
+}
+
+void SetDeviceMenuStyle() {
+
+  ImGuiStyle &style = ImGui::GetStyle();
+  style.Colors[ImGuiCol_Border] = {0.14f, 0.15f, 0.17f, 1.0f};
```
R-Abbasi, line 172: Since it's only few lines, you could simply use `ImGui::PushStyleColor` for that child window, I assume.

src/main.cpp
AKMaily: I am sure we will change them often and it is needed, esspecially for designing and including the different companys that are working on this project.

src/style.cpp
AKMaily: Right now it is only a few lines but 1. the style of this window will grow with time. 2. for readability and an easy access it is better to have this in a extra function .

R-Abbasi: It's rather weird to me how a free function with just one line of code increases readability and ease of access. 
Please note the difference between `ImGui::PushStyleColor`/`ImGui::PopStyleColor()` and `style.Colors`.

src/style.cpp
```diff
@@ -141,7 +141,7 @@ void SetMainWindowStyle() {
   style.Colors[ImGuiCol_Text] = {0.0f, 0.0f, 0.0f, 1.0f};
   style.Colors[ImGuiCol_WindowBg] = {1.0f, 1.0f, 0.93f, 1.0f};
   style.Colors[ImGuiCol_ChildBg] =
-      ImVec4(255 / 255.0f, 255 / 255.0f, 255 / 255.0f, 100 / 100.0f);
+     ImVec4(1.0f, 1.0f, 1.0f, 1.0f);
```
R-Abbasi, line 144: `ImVec4` is not required as it can be deduced. So a step towards more simplicity could be, e.g., `style.Colors[ImGuiCol_ChildBg] = {1.f, 1.f, 1.f, 1.f};`

src/style.cpp
R-Abbasi: You could probably also take the `std::vector` variables to the related header.

src/main.cpp
```diff
@@ -186,7 +186,7 @@ int main() {
     SetMainWindowStyle();
     ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, windowSize.x * .011f);
     ImGui::BeginChild("Record Data",
-                      {windowSize.x * 0.76f, windowSize.y * 0.65f},
+                      {windowSize.x * 0.76f, windowSize.y * 0.66f},
```
R-Abbasi, line 189: Not yet apparently: 


It looked fine already. Why did you change that too, please?

src/style.cpp
AKMaily: The design of the device menu is currently changed by Daniel. To implement the design described in the Omni-Vision document he will need more than one PushStyleColor/PopStyleColor function, therefor this function will have more than a few lines of code. I wrote this function so it is easier for him to know where he can change the design and to have a good readability even if the style is changed a lot.

src/main.cpp
```diff
@@ -184,7 +184,7 @@ int main() {
     // ############################ addPlots("Recording the data", ...)
     ImGui::Dummy({0.f, windowSize.y * .01f});
     SetMainWindowStyle();
-    ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, windowSize.x * .011f);
+    ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, windowSize.x * .009f);
```
R-Abbasi, line 187: I'd suggest `ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, windowSize.x * .01f);` and `io.FontGlobalScale = 1.35f;`

AKMaily: Why do you suggest changing the FontGlobalScale ?

src/style.cpp
```diff
@@ -97,6 +132,43 @@ void set_button_style_to(const nlohmann::json &config,
       ImVec4(load_json<Color>(config, "button", name, "active")));
 }
 
+void SetMainWindowStyle() {
```
R-Abbasi, line 137: Why didn't you set these settings inside the already existing function `SetupImGuiStyle`, please?

src/style.cpp
R-Abbasi: Have not used `ColorEdit*` already but managed this simple test: 
```
static float color[]{.3f, .2f, .1f, .7f};
ImGui::ColorEdit4("test", color);
ImGui::PushStyleColor(ImGuiCol_Button,
                      {color[0], color[1], color[2], color[3]});
ImGui::Button("Test color_1");
ImGui::Button("Test color_2");
ImGui::PopStyleColor();
``` 

I believe you can simplify your code this way.

src/style.cpp
AKMaily: Because i only want them to affect the main window and not any other region

src/main.cpp
R-Abbasi: You changed that to `1.2f` but it looks too small. `1.35f` seems better, I guess.

src/style.cpp
R-Abbasi: They're both for the main/parent window color settings, I guess. 
This supposedly makes the settings in the first function disabled as soon as the control reaches the second function in each frame, making the first function almost useless. 
For example, you've used `style.Colors[ImGuiCol_Text]` in both functions:
` style.Colors[ImGuiCol_Text] = ImVec4(1.0f, 1.0f, 1.0f, 1.0f);` 
` style.Colors[ImGuiCol_Text] = {0.0f, 0.0f, 0.0f, 1.0f};`

src/style.cpp
AKMaily: Currently trying to use a referenz for the color pointer, it seems to be a more complicated issue than thought, do you know how to fix this problem ?: 

´´´

  // Hier füge deinen Color Picker hinzu
  std::vector<ImVec4> copiedColors{};
  copiedColors.resize(ImGuiCol_COUNT); 

  static ImVec4 plotcolors[ImPlotCol_COUNT];
  // Color pickers for individual style elements
  if (ImGui::TreeNode("Style Colors")) {
    UpdateColors(styleElements, copiedColors);
    // Add more color pickers for other style elements

    ImGui::TreePop();
  }

  if (ImGui::TreeNode("ImPlotColors")) {
    UpdateColorsImPlot(plotElements, plotcolors);
    // Add more color pickers for other style elements

    ImGui::TreePop();
  }

  ImGui::End();
}

void UpdateColors(const std::vector<StyleElement> &styleElements,
                 std::vector<ImVec4> &colors) {

  ImGuiStyle &style = ImGui::GetStyle();
  int i = 0;
  for (const auto &element : styleElements) {
    ImGui::ColorEdit4(element.name, colors.data()[i]);
    ImVec4 colorWithAlpha = ImVec4(colors[i].x, colors[i].y, colors[i].z,
                                   style.Colors[element.colorIndex].w);
    style.Colors[element.colorIndex] = colorWithAlpha;
    i++;
  }
}

´´´

R-Abbasi: I'd suggest these:
1) Remove the functions `PopupStyleEditor`, `UpdateColors` and `UpdateColorsImPlot` from _header.hpp_
2) Declare the function `void PopupStyleEditor();` and define `std::vector`s below in that header: 
```
inline std::vector<StyleElement> colors = {
    {"Button", {0.145f, 0.156f, 0.168f, 1.f}, ImGuiCol_Button},
    {"Text", {1.f, 1.f, 1.f, 1.0}, ImGuiCol_Text}
    // rest
 };
inline std::vector<StyleElement> plotColors = {
    {"Plot Border", {1.f, 1.f, 1.f, 1.0f}, ImPlotCol_PlotBorder},
    {"Plot Axis Text", {1.f, 1.f, 1.f, 1.0f}, ImPlotCol_AxisText}
    // rest
};
```
3) And in `style.cpp`: 
```
void PopupStyleEditor() {
  ImGuiStyle &style = ImGui::GetStyle();
  if (ImGui::TreeNode("Style Colors")) {
    for (size_t i = 0; i < colors.size(); i++) {
      ImGui::ColorEdit4(colors[i].name, (float *)&colors[i].color);
      style.Colors[colors[i].colorIndex] = colors[i].color;
    }
    ImGui::TreePop();
  }

  ImPlotStyle &plotStyle = ImPlot::GetStyle();
  if (ImGui::TreeNode("Plot Colors")) {
    for (size_t i = 0; i < plotColors.size(); i++) {
      ImGui::ColorEdit4(plotColors[i].name, (float *)&plotColors[i].color);
      plotStyle.Colors[plotColors[i].colorIndex] = plotColors[i].color;
    }
    ImGui::TreePop();
  }
}
```
That, as mentioned, looks better and you should have the output like:

src/style.cpp
AKMaily: The settings should only be for the PlotRegion/ data window not the whole window. I guess the name of the function is wrong, will change that.

src/style.cpp
AKMaily: Is there a way to set the colors in the <StyleElements> as style.Colors[ImGuiCol_SOMETHING] and not as an ImVec4 {a,b,c,d}. It seems complicated to use it with imVec4 because the colors need to be the same as in SetUpImGuiStyle()

src/style.cpp
R-Abbasi: It seems you're using `GetStyle()` everywhere and `PushStyleColor/PopStyleColor()` nowhere. Did you look up their differences and when to use each, please? 
https://github.com/ocornut/imgui/blob/docking/imgui.h#L2054

src/style.cpp
R-Abbasi: Not sure I got your question well, sorry. Do you mean the `style.Colors[colors[i].colorIndex] = colors[i].color;` section in the `PopupStyleEditor` function, or the `std::vector`s inside the header, please?

AKMaily: I mean the vectors inside the structure : 

```

inline std::vector<StyleElement> plotColors = {
    {"Plot Border", {1.f, 1.f, 1.f, 1.f}, ImPlotCol_PlotBorder},
    {"Plot Axis Text", {1.f, 1.f, 1.f, 1.f}, ImPlotCol_AxisText}

```

R-Abbasi: If you mean this way:
```
inline std::vector<StyleElement> plotColors = {
    {"Plot Border", ImPlot::GetStyle().Colors[ImPlotCol_PlotBorder],
  ImPlotCol_PlotBorder}
```
`GetStyle()` needs `ImGui::CreateContext()` apparently which is done in `ImGuiInstance.hpp` called inside `main.cpp` and namespace scope variables are initialized at program startup, before `main` runs, hence it can't be done inside the header, supposedly. 

By the way, this:
```
inline std::vector<StyleElement> plotColors = {
    {"Plot Border", {1.f, 1.f, 1.f, 1.f}, ImPlotCol_PlotBorder}
```
doesn't seem complicated to me. You just copy the `ImVec4` data and paste it like the simple way above.

AKMaily: Okay i will try that out. I suggest using this method instead of the ImVec4 because if we use the ImVec4 everytime something changes in the SetUpImGuiStyle it also has to be copied into the <StyleElements> initialization which is prone to errors

R-Abbasi: There seems to be at least three ways for doing this:
- copy the new `ImVec4` data into the related item in the header
- only use the vectors in the header even for the items inside `SetUpImGuiStyle` - that way you don't need copying anything 
- use a `std::map` for the items inside the header - more complicated but better to me

AKMaily: Okay, i implemented it the way you suggested. It does not save the values of the colors permanently. I tried several ways to save them but i cant find a way without using a pointer. Do you have an idea how we could save the colors permanently ?

R-Abbasi: What do you mean by permanently, please?
If you mean during the time using the program you may go for the second way suggested above.

src/main.cpp
AKMaily: This version and the old master version does not use the FontScale from the config, but the default fontscale. This was implemented because otherwise the font gets pixelated. 
I suggest we should implement different fontsizes through different Fonts instead of using a fixed font and only scaling it. But i think this is a new issue, because in the old version the fontsize was also fixed to the default value.

src/style.cpp
```diff
@@ -391,181 +370,36 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
 }
 
 // For Development
-void PopupStyleEditor(bool bStyleDark, float alpha, nlohmann::json &config) {
-  ImGui::Begin("Style Editor", NULL, ImGuiWindowFlags_AlwaysAutoResize);
-
+void PopupStyleEditor() {
   ImGuiStyle &style = ImGui::GetStyle();
+  static ImGuiStyle style2 = ImGui::GetStyle();
```
R-Abbasi, line 372: 1) The theme won't be permeant if you close the popup.
2) The lines `static ImGuiStyle style2 = ImGui::GetStyle();`, `static ImPlotStyle styleImPlot2 = ImPlot::GetStyle();`, `style = style2;` and `styleImPlot = styleImPlot2;` seem redundant.

src/style.cpp
```diff
@@ -391,181 +370,36 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
...
-  if (ImGui::TreeNode("ImPlotColors")) {
-    UpdateColorsImPlot(plotElements, plotcolors);
-    // Add more color pickers for other style elements
+  static std::vector<ImVec4> plotColors{};
+  static std::vector<ImVec4> colorVec{};
 
-    ImGui::TreePop();
+  for (auto &element : style.Colors) {
+    colorVec.push_back(element);
   }
```
R-Abbasi, line 382: The loop will seemingly be continually pushing elements at the back of the vector as long as the popup is open. This looks unsafe (running out of memory for devices with limited recourses), unnecessary (because only the first few elements are used for `ImGui::ColorEdit4`) and not so performant (since it does heap memory allocations several times - if not dozens).

src/style.cpp
```diff
@@ -391,181 +370,36 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
...
+  }
+  if (ImGui::TreeNode("Plot Colors")) {
+    for (size_t i = 0; i < ImPlotCol_COUNT; i++) {
+      ImGui::ColorEdit4(ImPlot::GetStyleColorName(i), (float *)&plotColors[i]);
+      styleImPlot2.Colors[i] = plotColors[i];
+    }
+    styleImPlot = styleImPlot2;
+    ImGui::TreePop();
   }
 }
```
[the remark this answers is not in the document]
AKMaily: Okay thank you, i tried your suggestion and it seems to be better code. Will change that and push the results.

src/style.cpp
```diff
@@ -116,34 +117,30 @@ void set_button_style_to(const nlohmann::json &config,
...
 
+}
+void PopPlotRegionColors(){
+  for(int i = 0; i < 5; i++){
+  ImGui::PopStyleColor();
+  } 
+  for(int i = 0; i < 7; i++){
+  ImPlot::PopStyleColor();
+  } 
+}
```
R-Abbasi, line 143: `ImGui::PopStyleColor()` can take an argument, so I suggest removing the function `PopPlotRegionColors()` then adding `ImGui::PopStyleColor(5);` and ` ImPlot::PopStyleColor(7);` in `main.cpp`, like:
  ```
ImGui::EndChild(); // end child Record Data
  ImGui::PopStyleVar();
  ImGui::PopStyleColor(5);
  ImPlot::PopStyleColor(7);
```

Comment (AKMaily) [reply src/style.cpp:143]: I would like to keep this in a function because of two reasons. 1. It is better for readability as other programmers can directly see that the PopStyleColors depend on the pushStyleColors. 2. It is less prone to errors as changing something in the PushPlotRegionColor function can affect the PopStyleColors and in this way it is clear that they depend on each other. I will change the for loop in the PopPlotRegionColors as you suggested.

Comment (R-Abbasi) [reply src/style.cpp:143]: Then the function `PopPlotRegionColors()` ends up having this:
```
ImGui::PopStyleColor(5);
ImPlot::PopStyleColor(7);
```
Having a function with just two lines code doesn't seem very good to me.

AKMaily: As the code for the Plotwindow expands, the probability increases that the PopStyleColor() functions might become separated or that comments could end up in incorrect positions. Consolidating them into a function offers the safest approach to maintain readability. Also we want to keep the main function as small as possible.

R-Abbasi: In general, you code based on the current circumstances not expected future ones that may or may not come true. 
Anyway, am going to approve this `pr` as it's now reached a good status finally.

R-Abbasi: Approved.

R-Abbasi (approved):

PR #132, opened 2024-04-08
103 enhancement positioning of the buttons in the save menu

DanielNowak98:
Added changes in the save menu.

### DIFF
--- a/src/look_up_saves.hpp
+++ b/src/look_up_saves.hpp
@@ -36,6 +36,13 @@ inline std::string getSubdirectoriesInFolder(nlohmann::json language,
 
     if (isTxtInptFieldsEmpty) {
       const std::string newcar = "New Car";
+      ImGui::Separator();
+      ImGui::NewLine();
+      ImGui::Text("Additional Information");
+      ImGui::NewLine();
+
+      ImGui::Combo("Known Cars", &selectedOption, vins,
+                   static_cast<int>(subdirectories.size()));
 
       ImGui::InputText("Measurement", scantype, IM_ARRAYSIZE(scantype));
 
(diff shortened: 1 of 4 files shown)


src/look_up_saves.hpp
```diff
@@ -48,29 +55,30 @@ inline std::string getSubdirectoriesInFolder(nlohmann::json language,
...
 
-      if (selectedOption == 0) {
-        ImGui::InputText("Fin/Vin", VIN, IM_ARRAYSIZE(VIN));
-        selectedFolder = VIN;
-      }
+      ImGui::Separator();
+      ImGui::Text("Additional Information");
+      ImGui::NewLine();
+      ImGui::InputText("Measurement", scantype, IM_ARRAYSIZE(scantype));
+      ImGui::InputText("Mileage", mileage, IM_ARRAYSIZE(mileage));
```
R-Abbasi, line 75: Supposedly you could use the languages interface also for the string literals above.

src/menuscpp/Style.cpp
```diff
@@ -130,4 +130,11 @@ void SetMainWindowStyle() {
   // colors when hovering and clicking the axes
   style.Colors[ImGuiCol_ButtonHovered] = { 0.94f, 0.94f, 0.94f, 1.0f };
   style.Colors[ImGuiCol_ButtonActive] = { 0.94f, 0.94f, 0.94f, 1.0f };
+}
+
+void SetHorizontalSepeareatorColours() {
+
+  ImGuiStyle& style = ImGui::GetStyle();
+  style.Colors[ImGuiCol_Separator] = ImVec4(1.0f, 0.0f, 0.0f, 1.0f);
```
R-Abbasi, line 138: Since it's only few lines, you could simply use `ImGui::PushStyleColor` for the widget, I assume.

R-Abbasi: As far as I know, the `Save` button should appear only when at least one device is selected/checked, not otherwise.

src/saves_popup.cpp
```diff
@@ -220,44 +220,43 @@ void saves_popup(nlohmann::json const &config, nlohmann::json const &language,
...
-ImGui::Separator();
-ImGui::NewLine();
-if (ImGui::Button(" back ")) {
+  ImGui::Separator();
+  ImGui::NewLine();
+  if (ImGui::Button(" back ")) {
     ImGui::CloseCurrentPopup();
-}
-ImGui::SameLine(ImGui::GetWindowWidth() - 100); // Ändern Sie 100 entsprechend Ihrer Anforderungen
-if (ImGui::Button(" save ")) {
```
R-Abbasi: As far as I know, the `Save` button should appear only when at least one device is selected/checked, not otherwise.

DanielNowak98: Okay, i will check this.

PR #138, opened 2024-04-09
changed language settings on different positions

DanielNowak98:
Changed Language Linkings on different Positions.

### DIFF
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -1,7 +1,8 @@
-#include <set>
-#include <implot.h>
 #include "handler.hpp"
 #include "get_from_github.hpp"
+#include "languages.hpp"
+#include <implot.h>
+#include <set>
 
 void addPlots(const char *name, const bool flagPaused,
               std::function<void(double)> axesSetup) {
@@ -126,7 +127,7 @@ void devicesList() {
 
   if (sampler.has_value()) {
     for (auto &device : sampler->sampleDevices)
-      doDevice(device.first, "Messung");
+      doDevice(device.first, appLanguage[Key::Measurement]);
   } else {
     for (auto &device : devices)
       doDevice(device, "Ready");
--- a/src/languages.hpp
+++ b/src/languages.hpp
@@ -7,7 +7,14 @@
 
 enum class Key {
   Menu,
+  ButtonExplain,
+  FontSize,
+  WindowWidth,
+  WindowHeight,
   LanOption,
+  Restore,
+  RecordData,
+  Recordingdata,
   English,
   Settings,
   Reset,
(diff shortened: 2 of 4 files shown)


src/languages.hpp
```diff
@@ -33,6 +40,10 @@ enum class Key {
   Start,
   Stop,
   Continue,
+  Cancel,
```
R-Abbasi, line 43: _Cancel_ was not added because by default _Back_ was used instead of that.

src/languages.hpp
```diff
@@ -73,14 +84,23 @@ enum class Key {
...
     {Key::Settings, "Settings"},
     {Key::Reset, "Reset"},
     {Key::Reset_q, "Reset?"},
     {Key::Measure_not_saved, "The measurement was not saved!\n"
                              "Would you like to save it before deleting it?\n"},
+    {Key::Restore, "Restore"},
+    {Key::Devicelist, "Devicelist"},
+    {Key::RecordData, "Record Data"},
+    {Key::Recordingdata, "Recording the data"},
+    {Key::Save, "Save"},
```
R-Abbasi, line 99: _Save_ exists already, I guess.

src/main.cpp
```diff
@@ -169,27 +168,31 @@ int main() {
     ImGui::Dummy({0.f, windowSize.y * .01f});
     ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, windowSize.x * .011f);
     ImGui::PushStyleColor(ImGuiCol_Border, {0.3f, 0.4f, 0.7f, 0.6f});
-    ImGui::BeginChild("Record Data", {0.f, windowSize.y * 0.5f},
+    ImGui::BeginChild(appLanguage[Key::RecordData], {0.f, windowSize.y * 0.5f},
```
R-Abbasi, line 171: `ImGui::BeginChild`'s _id_ is not normally visible, so can't be sure how modifying it helps.

src/languages.hpp
DanielNowak98: Ahh okay, thank you!

src/handler.cpp
```diff
@@ -1,7 +1,8 @@
-#include <set>
-#include <implot.h>
 #include "handler.hpp"
 #include "get_from_github.hpp"
+#include "languages.hpp"
+#include <implot.h>
+#include <set>
```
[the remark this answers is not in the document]
DanielNowak98: I think this was done by the clang formatter.

src/main.cpp
AKMaily: In the child window for the plot the title is visible and is currently only in english @R-Abbasi

R-Abbasi: Do you mean [this BeginChild](https://github.com/skunkforce/OmniView/blob/master/src/main.cpp#L172), please?

R-Abbasi: As well as, it's not a title but an ID. The title is set by `addPlots("Recording the data" ...`.

AKMaily: Okay got it. So the language has to be changed directly in the addPlots function and not in BeginChild. @DanielNowak98

PR #141, opened 2024-04-19
Np/update release

NP199:
It is possible now to set a tag with following workflow:
git tag -a <versionnumber> (set local tag)
git push origin <versionnumber>

Example:
git tag -a v0.9.0
git push origin v0.9.0 

Note the tag will be added to the commit hash from your current branch you are working in.


.github/workflows/build.yaml
```diff
@@ -69,17 +69,66 @@ jobs:
...
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: ${{ contains(matrix.os, 'windows' ) }}
         with:
           name: ${{ github.event.repository.name }}-windows
-          path: build/Release/OmniView.exe
+          path: |
+            build/Release/OmniView.exe
+            config/config.json
+            languages/Deutsch.json
```
R-Abbasi, line 89: We don't use the `Deutsch.json` file in the project anymore. Moreover, we'll perhaps remove `config.json` as well due to its necessity of Internet connection and overhead.

AKMaily: In some of the old files the Deutsch.json is still used. It will be deleted after all files are updated with the language map. The config still exists so for this PR it is still used i guess.

AKMaily (approved):
This works, the first version will be v0.9.0

PR #145, opened 2024-04-23
Settings menu without bugs

AKMaily:
Deleted all unecessary code in the settingspopup.hpp


src/settingspopup.hpp
```diff
@@ -33,71 +33,21 @@ static void popup_settings(nlohmann::json &config, nlohmann::json &language,
     newconfig["text"]["scale"] = fontscale;
   }
```
R-Abbasi, line 35: You could use:
 - `#ifndef ... #define` header guard rather than `#pragma once`.
 - `static nlohmann::json newconfig = config;` instead of
 ```
 if (newconfig == 0) {
    newconfig = config;
  }
```
Also using `fontscale` when not initialized (`static float fontscale;`) could lead to UB, I guess, as primitive types don't have ctros.

Comment (AKMaily) [reply src/settingspopup.hpp:35]: This is code that existed before and i did not write it, but i will try to improve it. Thank you.

File: src/settingspopup.hpp [hunk 2]
```diff
@@ -33,71 +33,21 @@ static void popup_settings(nlohmann::json &config, nlohmann::json &language,
...
-
-  ImGui::SameLine();
-  if (ImGui::Button("Y-")) {
-    ButtonSizeY -= 10.0f;
-  }
-
-  if (ImGui::Button(load_json<std::string>(language, "button", "save").c_str(),
-                    ImVec2(ButtonSizeX, ButtonSizeY))) {
+  if (ImGui::Button(
+          load_json<std::string>(language, "button", "save").c_str())) {
```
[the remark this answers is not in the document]
AKMaily: I now added the appLanguage to the buttons , refactored the code a bit and added the language in the settingsmenu directly.

src/style.cpp
```diff
@@ -329,22 +330,6 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
...
-      appLanguage = englishLan;
-      showSettings = false;
-    }
-    if (ImGui::Button(appLanguage[Key::German])) {
-      appLanguage = germanLan;
-      showSettings = false;
-    }
-    ImGui::TreePop();
-  }
-  if (showSettings && ImGui::Button(appLanguage[Key::Settings])) {
```
R-Abbasi, line 347: Why did you not use the existing feature of closing the `TreePop` when an option is chosen, please?

src/settingspopup.hpp
```diff
@@ -33,71 +35,31 @@ static void popup_settings(nlohmann::json &config, nlohmann::json &language,
...
   }
 
-  ImGui::SameLine();
-  if (ImGui::Button("Y-")) {
-    ButtonSizeY -= 10.0f;
-  }
-
-  if (ImGui::Button(load_json<std::string>(language, "button", "save").c_str(),
-                    ImVec2(ButtonSizeX, ButtonSizeY))) {
+  if (ImGui::Button(appLanguage[Key::Save])) {
```
R-Abbasi, line 48: It seems `Save` is to save scale settings only not the selected language as well. 
Have you tested that button to see if it works as expected, please?

src/style.cpp
R-Abbasi: This way the `TreePop` will remain open if it's opened a second time. That doesn't seem quite good. The feature implemented earlier is doing something nice. You could take more time to comprehend the existing code and change it only if it's necessary. Moreover, having a popup only for the settings button among all others on the list doesn't seem quite reasonable to me.

The `Save` button doesn't even save the scale properly, as I just tested.

AKMaily: This pop-up window was already present in the interface. I integrated the language options into this pop-up to avoid having a separate tree node for users to change settings (such as language options), which would result in a strange user flow with an additional settings pop-up. Consolidating everything into a settings menu is a common practice seen in many other GUIs, making it a reasonable approach. Additionally, I'll investigate why the scale is not being saved correctly, as it appears to be a pre-existing bug.

AKMaily: Later on the settings menu should contain: Adding a username, adding a workshop ID, settings the fontsize, setting the language. Therefore this should be in another menu and not only as options in a treenode

src/settingspopup.hpp
AKMaily: The 'Attitude' button should be replaced with the 'Settings' button. This change aligns with the structure commonly seen in GUIs, where various functionalities are grouped under a unified settings menu. Just as we have menus for generating and analyzing data, having a settings menu is a standard practice. Placing the language option within this settings menu is logical, as language preference is a settings parameter. The 'Save' button should accurately preserve the user's current settings configuration from the settings menu, including language preference and font size.

R-Abbasi: The `OpenPopup` apparently uses the name for the `BeginPopupModal` as kind of an ID to keep it open internally. When the language changes that name also changes making the previous one sort of dead/vanished subsequently. 

I'd suggest the following changes:
`min.cpp`:
```
if (open_settings) 
  ImGui::OpenPopup(settingstitle.c_str());
```

`settingspopup.hpp`:

```
#ifndef SETTINGS_H
#define SETTINGS_H

#include "jasonhandler.hpp"
#include "languages.hpp"
#include <fmt/format.h>
#include <imgui.h>
#include <nlohmann/json.hpp>
#include <nlohmann/json_fwd.hpp>

static void popup_settings(nlohmann::json &config, nlohmann::json &language,
                           std::string const &configpath, bool &open_settings) {

  static float tempfontscale;
  static auto tempLan = appLanguage;
  ImGui::Text(appLanguage[Key::SettingsText]);
  ImGui::NewLine();

  if (tempfontscale < load_json<float>(config, "text", "minscale"))
    tempfontscale = load_json<float>(config, "text", "minscale");

  ImGui::TextUnformatted(
      fmt::format("{} {:.1f}", appLanguage[Key::FontSize], tempfontscale)
          .c_str());
  ImGui::SameLine();

  ImGuiIO &io = ImGui::GetIO();
  if (ImGui::Button("+")) {
    tempfontscale += 0.1f;
    io.FontGlobalScale = tempfontscale;
  }
  ImGui::SameLine();

  if (ImGui::Button("-") && (tempfontscale > 1.0f)) {
    tempfontscale -= 0.1f;
    io.FontGlobalScale = tempfontscale;
  }

  if (ImGui::TreeNode(appLanguage[Key::LanOption])) {
    if (ImGui::Button(appLanguage[Key::English]) && englishLan != appLanguage)
      appLanguage = englishLan;
    if (ImGui::Button(appLanguage[Key::German]) && germanLan != appLanguage)
      appLanguage = germanLan;
    ImGui::TreePop();
  }

  if (ImGui::Button(appLanguage[Key::Save])) {
    config["text"]["scale"] = tempfontscale;
    tempLan = appLanguage;
    open_settings = false;
    ImGui::CloseCurrentPopup();
  }
  ImGui::SameLine();
  if (ImGui::Button(appLanguage[Key::Back])) {
    io.FontGlobalScale = config["text"]["scale"];
    tempfontscale = config["text"]["scale"];
    appLanguage = tempLan;
    open_settings = false;
    ImGui::CloseCurrentPopup();
  }
  ImGui::SameLine();
  if (ImGui::Button(appLanguage[Key::Reset])) {
    io.FontGlobalScale = config["text"]["minscale"];
    tempfontscale = config["text"]["minscale"];
    appLanguage = germanLan;
    tempLan = germanLan;
  }
}

#endif // SETTING_H
```
Tired to make the code appear somewhat better without going into much detail in that.

AKMaily: Doesnt this still closes the Popup when the language is switched after pressing the english or german button ?

R-Abbasi: It should keep the popup open regardless of language change, I believe. Did you use both parts into the two files correctly, please?

R-Abbasi: Here the tree node is still open even after selecting the language which looks less responsive, to me. It became closed in prior code.

R-Abbasi: This removes the little overhead of opening the popup more than once and inconsistency of possible different positions for the popups: 
```
static void popup_settings(nlohmann::json &config, nlohmann::json &language,
                           std::string const &configpath, int& title) {
 // ....
 if (ImGui::TreeNode(appLanguage[Key::LanOption])) {
  if (ImGui::Button(appLanguage[Key::English]) && englishLan != appLanguage) {
    appLanguage = englishLan;
    title = 0; // index zero is English
  }
  if (ImGui::Button(appLanguage[Key::German]) && germanLan != appLanguage) {
    appLanguage = germanLan;
    title = 1; // index one is German
  }
  ImGui::TreePop();
}
 // .....
```

`main.cpp`: 
```
// ############################ Settings Menu
static std::vector titles{
    (std::string)englishLan.find(Key::Settings)->second + "###",
    (std::string)germanLan.find(Key::Settings)->second + "###"};
static int title = 0;
if (open_settings) {
  ImGui::OpenPopup(titles[title].c_str());
  open_settings = false;
}
if (ImGui::BeginPopupModal(titles[title].c_str(), nullptr,
                           ImGuiWindowFlags_AlwaysAutoResize)) {
  ImGui::SetItemDefaultFocus();
  popup_settings(config, language, configpath, title);
  ImGui::EndPopup();
}
```
We could, of course, check the return value of `find` too.

src/handler.cpp
```diff
@@ -1,7 +1,7 @@
-#include <set>
-#include <implot.h>
 #include "handler.hpp"
 #include "get_from_github.hpp"
+#include <implot.h>
+#include <set>
```
R-Abbasi, line 4: _clang-format_ changes the `include`d headers order too, that may not be wanted. You could copy the list of those headers beforehand and paste it after applying the _clang-format_ to preserve the order and avoid unexpected results.

src/handler.cpp
```diff
@@ -141,13 +141,20 @@ void set_config(const std::string &configpath) {
...
     fmt::print("Did not find {}.\n Download from Github\n\r",
                appLanguage[Key::German]);
     update_language_from_github();
   }
 }
+
+void set_inital_config(nlohmann::json &config) {
+  ImGuiIO &io = ImGui::GetIO();
+  io.FontGlobalScale = config["text"]["scale"];
+  fmt::print("{}\n", config["text"]["scale"].dump());
```
R-Abbasi: The value type of `["text"]["scale"]` in the `config` file is simply `float` and not `nlohmann_json` for which we use `dump()`, apparently.

src/settingspopup.hpp
R-Abbasi: The executable [here](https://github.com/skunkforce/OmniView/actions/runs/9000517595) doesn't load on my Windows machine.

src/handler.cpp
AKMaily: I checked afterwards if the programm still works, the header order is fine in this case. But thank you for your suggestion

src/handler.cpp
AKMaily: I am sorry i forgot to delete this line it was only for debugging. dump() is used because print cant transform the float of the config for internal reasons.

src/settingspopup.hpp
R-Abbasi: Approved.

PS: You don't need those many `commit`s for each `pr`, in general. Instead, you could make the expected changes, test the project as many times as required to assure you everything is working correctly and then `commit` them. That way reviewing a `pr` will expectedly be easier and more manageable, I think.

R-Abbasi (approved):

Issue #128, opened 2024-04-04
Refactoring: Saving data


NP199:
I wanted to save a big chunck of data and the programm crashed. The measurment was like 3 hours long, which worked fine, but when i tried to save it the program chrashed.

I think it will be important to give the User somekind of progressbar or feedback that the data will be saved.

My first thought on this was to save the data chunckwise and delte the already written chunck from the data vector until its empty.
The file stream can append each chunck to the file. Since the vector shrinks in size it could also be an indicator for an progressbar or feedback

AKMaily:
I guess a solution would be to implement two features. The first one is  to create a temp folder in the OmniView directory where the data chunks are saved temporarily while the measurement is taken. If the user wants to save the data, the saved file in the temp folder will be moved to the saves folder or the path the user selected. If the user does not want to save the file or if the user finished saving it in another folder the temp folder will be deleted. The second feature should be a progress bar that the user can see while the file from the one folder is moved to the folder he selected.

R-Abbasi:
> I think it will be important to give the User somekind of progressbar or feedback that the data will be saved.

 The progress bar is a good idea but it can't be that helpful regarding the crash, I guess.

- Took a measurement for about 10 minutes and saved it, but Windows couldn't open the created `.csv` file. Saw about 60 million lines/values when opened with *Notepad*. Most of the values are repetitive (sort of a recurrent pattern). Not sure if it's required to keep them intact.  

- Values, as I've explored, are positive and don't exceed 255. Hence, we may be able to use [std::vector<std::pair<std::uint16_t, std::int16_t>> samples;](https://github.com/AI-Gruppe/ai_omniscope-v2-communication_sw/blob/e084bad75322f5b0853116806a830311e3f4b8ad/src/OmniscopeSampler.hpp#L18) like `std::vector<std::pair<std::uint16_t, std::uint8_t>> samples;` to occupy less memory. 
- As well as, not sure if the number of lines is entirely necessary. If not, `std::vector<std::uint8_t> samples;` seems much better. 

> My first thought on this was to save the data chunckwise and delte the already written chunck from the data vector until its empty.

It will make the issue worse, I assume.

AKMaily:
> * Took a measurement for about 10 minutes and saved it, but Windows couldn't open the created `.csv` file. Saw about 60 million lines/values when opened with _Notepad_. Most of the values are repetitive (sort of a recurrent pattern). Not sure if it's required to keep them intact.

[...]

PR #154, opened 2024-05-13
saving process and progress bar

R-Abbasi:
- improved saving process performance a lot
- implemented a progress bar in order to observe the process 
- fixed [the issue](https://support.microsoft.com/en-us/office/what-to-do-if-a-data-set-is-too-large-for-the-excel-grid-976e6a34-9756-48f4-828c-ca80b3d0e15c?ns=excel&version=21&syslcid=1033&uilcid=1033&appver=zxl210&helpid=xlmain11.chm655696&ui=en-us&rs=en-us&ad=us) of storing a great deal of data into a `.csv` file not to lose any data 
- code clean-up and enhancements 

Here's a test for a half an hour measurement with more than180 million y values. It took 9 seconds for saving on my machine: 

Untitled

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,4 +67,4 @@ target_include_directories(OmniView PUBLIC ${LIBUSB_INCLUDE_DIRS})
 target_link_libraries(OmniView PUBLIC ${LIBUSB_LIBRARIES})
 
 target_compile_definitions(OmniView PUBLIC BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT)
-target_compile_features(OmniView PUBLIC cxx_std_20)
+target_compile_features(OmniView PUBLIC cxx_std_23)
--- a/src/generateTrainingData.cpp
+++ b/src/generateTrainingData.cpp
@@ -100,7 +100,7 @@ void generateTrainingData(
 
     static std::string Measurement;
     // one extra space for '\0' character
-    static char VIN[18];
+    static char VIN[19];
     static std::string Mileage;
     static std::vector<double> file_measuring_vals; // measurement values
     static bool grayFields = false;
@@ -236,8 +236,7 @@ void generateTrainingData(
       s += data->EventChar;
       // strlen is updated when entered char passes the filter
       size_t indx = strlen(VIN);
-
-      if (indx < 17)
+      if (indx < 18)
         return !std::regex_match(
             s, chars_regex); // return 0 as passed for matched chars
       return 1;              // discard exceeding chars
--- a/src/generateTrainingData.cpp
+++ b/src/generateTrainingData.cpp
@@ -117,31 +117,31 @@ void generateTrainingData(
       if (!readfile.is_open())
         fmt::println("Failed to open file {}", filename);
       else {
-        // first and second lines of file
-        std::string first_line, second_line;
-        for (size_t i = 0; !readfile.eof(); i++) {
-          if (i == 0) {
-            std::getline(readfile, first_line);
-            first_line.pop_back(); // remove ending new line
-            std::stringstream ss(first_line);
-            // extract input fields data from the first line
-            for (size_t j = 0; j < fieldsSize; j++) {
-              std::string substr;
-              std::getline(ss, substr, ',');
-              FieldsData[j] = substr;
-            }
-          } else if (i == 1)
-            std::getline(readfile, second_line); // device ID
-          else {
-            // read measuring values into the vector
-            constexpr size_t bigNumber{10'000'000};
-            readfile.ignore(bigNumber, ',');
-            double value{};
-            readfile >> value;
-            file_measuring_vals.emplace_back(value);
-            // at the last loop, the last number is picked and loop goes on
-            // vector pushes value before eof is reached
-          }
+        // first, second and third lines of file
+        std::string first_line, second_line, third_line;
+        std::getline(readfile, first_line);
+        first_line.pop_back(); // remove ending new line
+        std::stringstream ss(first_line);
+        // extract input fields data from the first line
+        for (size_t j = 0; j < fieldsSize; j++) {
+          std::string substr;
+          std::getline(ss, substr, ',');
+          FieldsData[j] = substr;
+        }
+        std::getline(readfile, second_line); // device ID
+        second_line.pop_back();
+        std::getline(readfile, third_line); // samplying rate
+        third_line.pop_back();
+        while (!readfile.eof()) {
+          // read measuring values into the vector
+          double value{};
+          readfile >> value;
+          file_measuring_vals.emplace_back(value);
+          static constexpr size_t bigNumber{10'000'000};
+          readfile.ignore(bigNumber, ' '); // two spaces between elements
+          readfile.ignore(bigNumber, ' ');
+          // at the last loop, the last number is picked, loop
+          // goes on and vector pushes value before eof is reached
         }
         // pop the extra last element
         file_measuring_vals.pop_back();
--- a/src/languages.hpp
+++ b/src/languages.hpp
@@ -78,7 +78,8 @@ enum class Key {
   German,
   Ready,
   FontSize,
-  SettingsText
+  SettingsText,
+  saving
 };
 
 inline const std::map<Key, const char *> englishLan{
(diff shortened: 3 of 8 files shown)


NP199:
I have two issue. When the Popup is opened, the width of the popup is my main window width and not entirely visible anymore.
Screenshot WindowSize

And the csv file format is not correct. This is not an valid .csv file format.

Screenshot CSVFile

src/languages.hpp
```diff
@@ -233,7 +235,8 @@ inline const std::map<Key, const char *> germanLan{
     {Key::German, "Deutsch"},
     {Key::Ready, "Bereit"},
     {Key::FontSize, "Schriftgröße"},
-    {Key::SettingsText, "Legen sie hier ihre persönlichen Einstellungen fest"}};
+    {Key::SettingsText, "Legen sie hier ihre persönlichen Einstellungen fest"},
+    {Key::saving, "sparen ..."}};
```
NP199, line 239:
```suggestion
    {Key::saving, "speichern ..."}};
```

"sparen" means something like saving money in english.

src/saves_popup.cpp
```diff
@@ -10,34 +13,36 @@ namespace fs = std::filesystem;
...
-
-  for (std::size_t i{}; i < values.size(); ++i) {
-    fileContent += fmt::format("{},", i);
-    if (values[i].second)
-      fileContent += fmt::format("{}\n", values[i].second);
-  }
-
+                 const fs::path &outFile, std::string allData, size_t &y_indx) {
+
+  allData += fmt::format("\n{}-{}\nSamplying rate is 100000\n", device.type,
```
NP199, line 18:
```suggestion
  allData += fmt::format("\nType: {}\nSerial: {}\nSamplingRate: {}\n", device.type, device.serial, device.sampleRate);
```

R-Abbasi:
The first case belongs to [this line](https://github.com/skunkforce/OmniView/pull/154/files#diff-d8f5a1ea29ec3870a7acfc39166a6cfe765c1cf5b2cf6859f80deee7797da262R232) and should be fine. Here's a quick test on my machine: 


The second case is a remedy to fix an issue for which there's already a link in the description.

src/saves_popup.cpp
```diff
@@ -10,34 +13,36 @@ namespace fs = std::filesystem;
...
+                         device.serial);
+  std::string fileContent;
+  fileContent.resize_and_overwrite(
+      // five bytes for each y_value, three for the number
+      // and two spaces as a separator between the numbers
+      values.size() * 5, [&values, &y_indx](char *begin, std::size_t) {
+        auto end = begin;
+        for (; y_indx < values.size(); y_indx++) {
+          end = std::to_chars(end, end + 3, values[y_indx].second).ptr;
+          *end++ = ' ';
```
NP199, line 28:
```suggestion
          *end++ = ', ';
```

src/saves_popup.cpp
```diff
@@ -10,34 +13,36 @@ namespace fs = std::filesystem;
...
+          *end++ = ' ';
```
NP199, line 29:
```suggestion
          *end++ = ' \n';
```

src/generateTrainingData.cpp
NP199, line 1: This file is getting pretty huge. generateTrainigndata could get some seperate functions for specific tasks

src/saves_popup.cpp
```diff
@@ -1,5 +1,8 @@
 #include <fstream>
 #include <sstream>
+#include <future>
+#include <charconv>
+#include <format>
```
NP199, line 5: Why do we need the format header?

src/saves_popup.cpp
R-Abbasi: If storing each element on a line is fine then I suppose we can go simpler, like: 
```
values.size() * 4, [&values, &y_indx](char *begin, std::size_t) {
  auto end = begin;
  for (; y_indx < values.size(); y_indx++) {
    end = std::to_chars(end, end + 3, values[y_indx].second).ptr;
    *end++ = '\n';
```

src/saves_popup.cpp
R-Abbasi: To stick more with STL features instead of external libraries. 
Unfortunately GCC doesn't quite support `<print>`, we could probably otherwise remove `fmt` from the project totally.

CMakeLists.txt
```diff
@@ -67,4 +67,4 @@ target_include_directories(OmniView PUBLIC ${LIBUSB_INCLUDE_DIRS})
 target_link_libraries(OmniView PUBLIC ${LIBUSB_LIBRARIES})
 
 target_compile_definitions(OmniView PUBLIC BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT)
-target_compile_features(OmniView PUBLIC cxx_std_20)
+target_compile_features(OmniView PUBLIC cxx_std_23)
```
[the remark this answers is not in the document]
R-Abbasi: Read the second solution text [here](https://github.com/skunkforce/OmniView/issues/128#issuecomment-2089733347), please.

src/main.cpp
```diff
@@ -160,14 +155,19 @@ int main() {
...
-        (std::string)germanLan.find(Key::Settings)->second + "###ID"};
+    // ############################ Settings Menu
     static int title = 0;
+    static std::vector<std::string> titles(2); // two languages
     if (open_settings) {
-      ImGui::OpenPopup(titles[title].c_str());
+      const auto EngItr = englishLan.find(Key::Settings);
+      const auto GrmItr = germanLan.find(Key::Settings);
+      // check returned value from find() and set titles 
+      if (EngItr != englishLan.end() && GrmItr != germanLan.end()) {
```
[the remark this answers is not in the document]
R-Abbasi: Did you read the comment right above it, please?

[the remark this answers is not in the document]
AKMaily: Yes and i dont understand why we need to check the value from find() ?

src/saves_popup.cpp
NP199: Sounds good to me

CMakeLists.txt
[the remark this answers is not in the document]
AKMaily: Is there a possible solution without using c++23 ?

src/saves_popup.cpp
NP199: But do we use features from <format> right now?

src/saves_popup.cpp
R-Abbasi: But then you can save only a limited (precious few) number of values into the file - so that it can be opened without losing data - and have measurements for merely less than one minute, supposedly.

src/saves_popup.cpp
R-Abbasi: We can omit using `<format>` for now; it's used on one line solely, but I assume, the more important approach to consider is to go with new features and updated versions of compilers to use them.

src/main.cpp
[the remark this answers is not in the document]
R-Abbasi: Read [this](https://en.cppreference.com/w/cpp/container/map/find), please.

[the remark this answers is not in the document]
AKMaily: In which case are the settings values not found ?

NP199:
@R-Abbasi please check if gcc 13 compiles your features. Then we can use this workflow fix https://github.com/skunkforce/OmniView/tree/np/GCCTest

In there is an update to the gcc version 13 for Ubuntu, this could help building the new c++23 features.

R-Abbasi:
That's good news.
Sure [it works](https://godbolt.org/z/G7bP9E9fE) with gcc 13.

NP199:
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md
Here different software version which can be used.
The beta Ubuntu2404 version could be used for gcc version 14. 
But i think we should wait for the release version

R-Abbasi:
I suggest that we go for GCC 13 for now, as it works, and use the `resize_and_overwrite` API. We may consider using `<print>` later on when there's a CI runner on GitHub for GCC 14. 
Do you agree, please?

NP199:
I agree

AKMaily:
The .csv file needs to be in a valid format. This includes 1. new lines between the different data points 2. a header where the metadata is saved in only one line

R-Abbasi:
As for 2, what do you mean by that metadata, please?
We now have string literals of the input text fields on the first line, device type on the second, device serial on the third and sampling rate on the fourth. Do you mean to have these all on one line, please?

AKMaily:
> As for 2, what do you mean by that metadata, please? We now have string literals of the input text fields on the first line, device type on the second, device serial on the third and sampling rate on the fourth. Do you mean to have these all on one line, please?

yes they should be all in one line, seperated by a comma

AKMaily (approved):

PR #156, opened 2024-05-22
loading old data and a few code refinements

R-Abbasi:
- coded loading old data and displaying it onto the plot region 
- managed the required menu 
- added a number of code refinements 

PS: This is the first step to the related issue as it's not yet completely clear how the implementation and design are going to be for the feature.

### DIFF
--- a/src/languages.hpp
+++ b/src/languages.hpp
@@ -239,4 +239,5 @@ inline const std::map<Key, const char *> germanLan{
     {Key::Saving, "speichern ..."}};
 
 inline auto appLanguage = englishLan;
+namespace fs = std::filesystem;
 #endif
--- a/src/languages.hpp
+++ b/src/languages.hpp
@@ -79,7 +79,8 @@ enum class Key {
   Ready,
   FontSize,
   SettingsText,
-  Saving
+  Saving,
+  Load_file_Data
 };
 
 inline const std::map<Key, const char *> englishLan{
@@ -157,7 +158,8 @@ inline const std::map<Key, const char *> englishLan{
     {Key::Ready, "Ready"},
     {Key::FontSize, "Fontsize"},
     {Key::SettingsText, "Set your personal settings for the software"},
-    {Key::Saving, "saving ..."}};
+    {Key::Saving, "saving ..."},
+    {Key::Load_file_Data, "Load file data"}};
 
 
 inline const std::map<Key, const char *> germanLan{
@@ -236,7 +238,8 @@ inline const std::map<Key, const char *> germanLan{
     {Key::Ready, "Bereit"},
     {Key::FontSize, "Schriftgröße"},
     {Key::SettingsText, "Legen sie hier ihre persönlichen Einstellungen fest"},
-    {Key::Saving, "speichern ..."}};
+    {Key::Saving, "speichern ..."},
+    {Key::Load_file_Data, "Dateidaten laden"}};
 
 inline auto appLanguage = englishLan;
 namespace fs = std::filesystem;
--- a/src/languages.hpp
+++ b/src/languages.hpp
@@ -239,7 +239,7 @@ inline const std::map<Key, const char *> germanLan{
     {Key::FontSize, "Schriftgröße"},
     {Key::SettingsText, "Legen sie hier ihre persönlichen Einstellungen fest"},
     {Key::Saving, "speichern ..."},
-    {Key::Load_file_Data, "Dateidaten laden"}};
+    {Key::Load_file_Data, "Alte Daten laden"}};
 
 inline auto appLanguage = englishLan;
 namespace fs = std::filesystem;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -15,14 +15,9 @@ int main() {
   auto now = std::chrono::system_clock::now();
   std::time_t now_time_t = std::chrono::system_clock::to_time_t(now);
   std::tm now_tm = *std::gmtime(&now_time_t);
-  bool open_settings = false;
-  bool open_generate_training_data = false;
-  static bool flagPaused = true;
-  bool flagDataNotSaved = true;
-  bool Development = false;
-  bool flagInitState = true;
-  bool load_old_data = false;
-  bool load_old_data_visible = true;
+  static bool flagPaused{true};
+  bool flagDataNotSaved{true}, Development{false}, flagInitState{true},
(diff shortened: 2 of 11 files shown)


AKMaily:
Pressing the reset button in the software should not reset the loaded data.

src/languages.hpp
```diff
@@ -236,7 +238,9 @@ inline const std::map<Key, const char *> germanLan{
     {Key::Ready, "Bereit"},
     {Key::FontSize, "Schriftgröße"},
     {Key::SettingsText, "Legen sie hier ihre persönlichen Einstellungen fest"},
-    {Key::Saving, "speichern ..."}};
+    {Key::Saving, "speichern ..."},
+    {Key::Load_file_Data, "Dateidaten laden"}};
```
AKMaily, line 242: "Alte Daten laden"

src/style.cpp
```diff
@@ -329,14 +340,49 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
...
+std::optional<Omniscope::Id>
+load_file(fs::path &loadedFileName) { // load and display data from file
+  std::ifstream readfile(loadedFileName, std::ios::binary);
+  if (!readfile.is_open())
+    fmt::println("Failed to open file {}", loadedFileName);
+  else {
+    size_t indx{2};
+    std::string first_line;
+    std::getline(readfile, first_line);
+    std::istringstream ss{first_line};
```
AKMaily, line 358: please give this variable a better defined name instead of ss

src/style.cpp
```diff
@@ -329,14 +340,49 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
...
+      if (i == 3) // fourth element (Type of scope)
+        id.type = substr;
+      if (i == 4) // fifth element (serial of scope)
+        id.serial = substr;
+    }
+    while (!readfile.eof()) {
+      double value{};
+      readfile >> value;
+      captureData[id].emplace_back(indx++, value);
+      static constexpr size_t bigNumber{10'000'000};
```
AKMaily, line 374: what will happen if there are more than 10 million values?

AKMaily (commented):
Loading the data works fine, although it should not be deleted when the user resets the measurement. There should be an extra reset button in the "Found Devices" section where the user can reset all loaded data.

R-Abbasi:
> Pressing the reset button in the software should not reset the loaded data.

It wasn't mentioned in the task description. 
Could you explain that a little more, please?

R-Abbasi:
> Loading the data works fine, although it should not be deleted when the user resets the measurement. There should be an extra reset button in the "Found Devices" section where the user can reset all loaded data.

Resetting files loaded data is done using the checkbox. A file's data is loaded and displayed as soon as the checkbox is checked and erased/reset otherwise.

src/languages.hpp
```diff
@@ -236,7 +238,9 @@ inline const std::map<Key, const char *> germanLan{
...
 
 inline auto appLanguage = englishLan;
+namespace fs = std::filesystem;
```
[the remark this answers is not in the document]
R-Abbasi: I agree that it may appear a little odd but the header was the only one `include`d where `std::filesystem` is utilized in TUs. The other way is to have the definition more than once but I think for the time being it's fine.

src/main.cpp
```diff
@@ -187,14 +188,13 @@ int main() {
...
 
-    addPlots("Recording the data", flagPaused, [&xmax_paused](double x_max) {
+    addPlots("Recording the data", [](double x_max) {
       if (!flagPaused) {
         ImPlot::SetupAxes("x [Data points]", "y [ADC Value]",
                           ImPlotAxisFlags_AutoFit, ImPlotAxisFlags_AutoFit);
         ImPlot::SetupAxisLimits(ImAxis_X1, x_max - 7500, x_max + 7500,
                                 ImGuiCond_Always);
       } else {
-        xmax_paused = x_max;
```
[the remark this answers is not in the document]
R-Abbasi: It doesn't seem to have any usage.

src/style.cpp
R-Abbasi: `input`

src/style.cpp
R-Abbasi: It extracts characters from the input sequence and discards them, until either `bigNumber` characters have been extracted, or one compares equal to `\n`. Have a look at [the Docs](https://en.cppreference.com/w/cpp/io/basic_istream/ignore), please.

AKMaily:
> > Loading the data works fine, although it should not be deleted when the user resets the measurement. There should be an extra reset button in the "Found Devices" section where the user can reset all loaded data.
> 
> Resetting files loaded data is done using the checkbox. A file's data is loaded and displayed as soon as the checkbox is checked and erased/reset otherwise.

Yes the files are reset using the checkboxes which is correct. But they are also reset when the user has the following userflow: load old data--> measure new data with a scope (press play button)-->press reset button. In this case not only the measured data is reset but also all loaded files are reset which should not happen.

src/style.cpp
AKMaily: The variable name should explain the usage of the variable

src/main.cpp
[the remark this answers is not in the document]
AKMaily: Do you know why this was implemented this way @NP199 ?

src/style.cpp
R-Abbasi: Identities naming is not as easy as you expect and abbreviations are used commonly. To me both `ss` or `input` are fine.

R-Abbasi:
Managed something to still have the data shown even after pressing the reset button which is rather suitable for the first step.

AKMaily:
> Managed something to still have the data shown even after pressing the reset button which is rather suitable for the first step.

If you are finished with this version please commit it so we can test and review it

R-Abbasi:
It's already implemented in the current `commit`.

AKMaily:
> It's already implemented in the current `commit`.

Okay great, i tested that. Not reseting the data with the Reset button now works, but if you try the following user flow: loading old data, not checking the box, search for devices --> start measurement --> reset measurement. The loaded data is displayed in the window even when it is not selected : 
image

R-Abbasi:
Fixed the bug.

AKMaily:
> Fixed the bug.

That is great, please also add the reset button for the found devices region so old loaded measurements can be reset

PR #158, opened 2024-05-24
119 enhancement toolbar region

AKMaily:
Added the toolbar buttons as Icons.


src/imagesHeaderToolbar.hpp
```diff
@@ -0,0 +1,432 @@
+// imagesHeader.hpp
```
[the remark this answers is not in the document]
AKMaily: I added this header file to keep the different regions seperated. Also i got an error because the imagesHeader.hpp had to be included in the main.cpp and the style.cpp. Is there a better way to include this ?

src/main.cpp
```diff
@@ -83,44 +85,105 @@ int main() {
...
-          if (ImGui::Button(appLanguage[Key::Start], toolBtnSize)) {
+          if (ImGui::ImageButton(
+                  appLanguage[Key::Start],
+                  (void *)(intptr_t)image_texture[PngRenderedCnt],
+                  ImVec2(image_width[PngRenderedCnt] * iconsSacle,
+                         image_height[PngRenderedCnt] * iconsSacle))) {
             sampler.emplace(deviceManager, std::move(devices));
             flagPaused = false;
             flagDataNotSaved = true;
+            for (auto &device : devices) {
```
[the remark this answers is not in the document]
AKMaily: This code is from an older version. It turns out we deleted it in the master version, but it is needed to stop the LED on the oscilloscope from blinking when the measurement is stopped.

src/imagesHeaderToolbar.hpp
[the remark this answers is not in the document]
R-Abbasi: I think we need to keep our header files in accordance with what has been ended up in the _OmniView-Product-Vision and don't extend their number. 
You could perhaps do the implementation in the `style.cpp` where the images header is already included.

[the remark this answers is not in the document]
AKMaily: Implementing the toolbar and its functionality is not part of the style.cpp, as it does not belong to the style or the SideBarMenu. In the OmniView-Product-Vision, the relevant files include not only the style.cpp but also the cpp files for different regions: SideBar, Devices Found, Plot Region, and Toolbar. I could create a cpp file to handle the toolbar, but I suspect this won't solve the problem. Using a separate header file for the images in the different categories doesn't seem too complicated. Additionally, we could place all the images in one folder named "assets" for better organization.

src/main.cpp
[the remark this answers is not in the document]
R-Abbasi: That makes sense for the following uses but here it shouldn't do anything because as mentioned `devices` has been `std::move`d in advance.

src/imagesHeaderToolbar.hpp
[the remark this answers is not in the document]
R-Abbasi: Still think we ought not to add a new header for potentially every feature it could be used in to void leading codebase to result in kind of the messy situation it used to be for some time. `style.hpp/cpp` appear to pertain to style, hence the name.

src/main.cpp
```diff
@@ -83,44 +85,105 @@ int main() {
...
       set_button_style_to(config, "stop");
-      if (ImGui::Button(appLanguage[Key::Stop], toolBtnSize))
+      if (ImGui::ImageButton(
+              appLanguage[Key::Stop],
+              (void *)(intptr_t)image_texture[PngRenderedCnt],
+              ImVec2(image_width[PngRenderedCnt] * iconsSacle,
+                     image_height[PngRenderedCnt] * iconsSacle))) {
         flagPaused = true;
+        for (auto &device : sampler->sampleDevices) {
+          device.first->send(Omniscope::Stop{});
```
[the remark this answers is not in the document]
NP199: Do you mean the sampleDevies are moved beforehand?

src/style.cpp
```diff
@@ -335,6 +336,181 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
                   .c_str());
 }
 
+void set_toolbar(const nlohmann::json &config, const nlohmann::json &language, bool &flagPaused){
+
+  // variable declaration
```
R-Abbasi, line 341: These definitions seem to be repeated per frame.

src/style.cpp
```diff
@@ -335,6 +336,181 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
                   .c_str());
 }
 
+void set_toolbar(const nlohmann::json &config, const nlohmann::json &language, bool &flagPaused){
```
R-Abbasi, line 339: `flagPaused` is not used in the function.

src/style.cpp
```diff
@@ -335,6 +336,181 @@ void set_side_menu(const nlohmann::json &config, bool &flagPaused,
...
+        set_button_style_to(config, "stop");
+        if (ImGui::ImageButton(
+                appLanguage[Key::Reset],
+                (void *)(intptr_t)image_texture[PngRenderedCnt],
+                ImVec2(image_width[PngRenderedCnt] * iconsSacle,
+                       image_height[PngRenderedCnt] * iconsSacle))) {
+          if (flagDataNotSaved) {
+            ImGui::OpenPopup(appLanguage[Key::Reset_q]);
+            for (auto &device : sampler->sampleDevices) {
+              device.first->send(Omniscope::Stop{});
```
R-Abbasi, line 467: It looks redundant.

src/style.cpp
AKMaily: line 402

src/style.cpp
AKMaily: Redundant to which part ?

src/style.cpp
R-Abbasi: If you take a look at the [master branch](https://github.com/skunkforce/OmniView/blob/master/src/main.cpp#L14) it'll be obvious they're defined before the main loop.

src/style.cpp
R-Abbasi: Sorry, I meant `flagPaused` in `set_side_menu`.
Have already fixed it in my current `pr`.

src/style.cpp
R-Abbasi: I meant, the loop is required seemingly for the two buttons below merely, not others: `appLanguage[Key::Stop]` for which `device.first->send(Omniscope::Stop{});` and `appLanguage[Key::Continue]` for which `device.first->send(Omniscope::Star{});` to be used, respectively.

src/style.cpp
AKMaily: Okay then we will fix that we your current PR instead of mine

src/style.cpp
AKMaily: okay i will make them static

src/style.cpp
```diff
@@ -389,7 +389,7 @@ void set_toolbar(const nlohmann::json &config, const nlohmann::json &language, b
     static int image_height[size];
     static int image_width[size];
     static GLuint image_texture[size];
-    static constexpr float iconsSacle{0.8f};
+    static constexpr float iconsSacle{0.8};
```
R-Abbasi, line 392: Why did you delete `f` at the end of the number, please?
`0.8` is `double` and that way an implicit conversion is to happen.

R-Abbasi:
There're still some inconsistencies concerning both toolbar buttons' resizing and the `Save` button default theme. 

Capture

AKMaily:
> There're still some inconsistencies concerning both toolbar buttons' resizing and the `Save` button default theme.
> 
> Capture

Resizing of the toolbar and colors will be fixed by @tesourinha in a second PR

src/style.cpp
R-Abbasi: `void set_toolbar (...)` looks too long.

src/style.cpp
```diff
@@ -347,6 +348,178 @@ void set_side_menu(const nlohmann::json &config, bool &open_settings,
...
+
+    // The order matters because of the counter for the images !!!
+    static const unsigned char *imagesNames[] = {
+        PlayButton_png, PauseButton_png, SaveButton_png, ResetButton_png};
+    static const unsigned int imagesLen[] = {
+        PlayButton_png_len, PauseButton_png_len, SaveButton_png_len,
+        ResetButton_png_len};
+    // Load the images for the SideBarMenu
+    for (int i = 0; i < size; i++)
```
R-Abbasi, line 401: You could also use a `size_t` type for `i` to match `size`'s.

AKMaily: What exactly is too long about it ?

src/main.cpp
```diff
@@ -1,7 +1,9 @@
+
```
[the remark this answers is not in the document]
AKMaily: Ran clang-formatter over it, could be because of that

src/style.cpp
AKMaily: Will change that again

src/style.cpp
R-Abbasi: The function itself.

R-Abbasi:
Approved. 

PS: The `pr` yet has some inconsistencies and bugs which need to be addressed in the next `pr`s.

PR #159, opened 2024-05-29
149 id in settingsmenu

tesourinha:
To compile, you have to use the config that is in the branch, make sure the key-value pair  `"Customer_Id": ""` exists and set the path to this config in the settingspopup.hpp. In the master version, the key-value pair `"Customer_Id": ""` needs to be added.

### DIFF
--- a/src/settingspopup.hpp
+++ b/src/settingspopup.hpp
@@ -9,7 +9,7 @@
 #include <nlohmann/json.hpp>
 #include <nlohmann/json_fwd.hpp>
 
-static void popup_settings(nlohmann::json &config, nlohmann::json &language,
+static void popup_settings(nlohmann::json &config,
                            std::string const &configpath, int &title) {
 
   static float tempfontscale = config["text"]["scale"];
--- a/src/settingspopup.hpp
+++ b/src/settingspopup.hpp
@@ -5,9 +5,9 @@
 #include "languages.hpp"
 #include <fmt/format.h>
 #include <imgui.h>
+#include <imgui_stdlib.h>
 #include <nlohmann/json.hpp>
 #include <nlohmann/json_fwd.hpp>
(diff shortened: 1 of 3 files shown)


R-Abbasi:
The program produces a crash when `Attitude` is clicked.

src/settingspopup.hpp
```diff
@@ -20,6 +21,47 @@ static void popup_settings(nlohmann::json &config,
   ImGui::Text(appLanguage[Key::SettingsText]);
   ImGui::Text("                            ");
 
+  // Create and initialize customer_id variable
+  static char customer_id[20] = "";
```
R-Abbasi, line 25: `static` primitive data types are zero initialized by default. So you could define them like `static char customer_id[20];`, `static bool isEmpty;` and so forth. Of course being explicit here can be fine. 
Moreover, you could have the size `size_t sz/size {20};` and use it for the array and others if needed.

src/settingspopup.hpp
```diff
@@ -20,6 +21,47 @@ static void popup_settings(nlohmann::json &config,
...
+  static bool isempty = false;
+
+  // Initialize customer_id only once
+  if (!is_initialized) {
+    std::string temp = config["Costumer_id"].get<std::string>();
+    std::copy(temp.begin(), temp.end(), customer_id);
+    if (temp.empty()) {
+      isempty = true;
+    }
+    customer_id[temp.size()] = '\0'; // Ensure null termination
```
R-Abbasi, line 39: Don't you need any bound checking before using the array here, please?

AKMaily:
> The program produces a crash when `Attitude` is clicked.

this is normal as the config is downloaded from the master in this version, which hasnt the "customer_id" field. This should be fixed when the branch is included in the master. For testing please manually load the config from this branch.

PR #162, opened 2024-06-07
Feature loading old data second iteration

R-Abbasi:
- Made changes into `header.hpp`, `header.cpp`, `imagesHeader.hpp`, `languages.hpp`, `main.cpp`, `style.cpp` and `style.hpp` to implement the feature mentioned in the _Loading old data, second iteration #160_ issue. 
- code cleanup, refactoring and improvements


AKMaily (commented):
Only viewed the handler.cpp a bit and tested the last version. Found a bug that applies to many userflows. After loading the first file, this file seems to be saved and no other file can be loaded. Even after reseting the old file and loading a new one, the old file will always be loaded.

AKMaily:
Testing the application a second bug occures. When taking a measurement and also having an old file selected, this file is listed as a device in the saving menu. 
image

src/languages.hpp
```diff
@@ -239,7 +245,10 @@ inline const std::map<Key, const char *> germanLan{
     {Key::FontSize, "Schriftgröße"},
     {Key::SettingsText, "Legen sie hier ihre persönlichen Einstellungen fest"},
     {Key::Saving, "speichern ..."},
-    {Key::Load_file_Data, "Alte Daten laden"}};
+    {Key::Load_file, "Datei laden"},
+    {Key::Load_file_data, "Alte Daten laden"},
+    {Key::Load_another_file, "Eine andere Datei laden"},
```
AKMaily, line 250: "Eine weitere Datei laden "

src/handler.cpp
```diff
@@ -131,6 +133,82 @@ void devicesList() {
       doDevice(device, appLanguage[Key::Ready]);
 }
 
+void load_files(decltype(captureData) &loadedDvcs,
+                std::map<Omniscope::Id, std::string> &Dvcs_filenames, bool& loadFile) {
+  static std::set<fs::path> loadedFileNames;
+  auto do_load = [&loadedDvcs, &Dvcs_filenames] {
```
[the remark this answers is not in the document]
R-Abbasi: I think it's not a bug because loading files and handling them work based on the ID of the files and many files with the same ID are added only once. That could change though.

src/languages.hpp
R-Abbasi: That's because the loaded file is added into the `captureData` to be shown on the plot region which is not a bug by itself.  But will try to manage the saving section as well.

src/handler.cpp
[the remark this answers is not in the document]
AKMaily: Yes, that should also change. It should depend on the filename rather than the device ID. This issue also occurs in the following user flow: load a file → display the file → reset the file via the reset button → load another file. In this scenario, the new file should be loaded since the previous file was reset, but currently, the old file is loaded again.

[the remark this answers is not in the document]
R-Abbasi: No. It's not the way `std::map` works. It needs unique IDs. Have a look at [Docs for `std::map`](https://en.cppreference.com/w/cpp/container/map), please. A device (a scope) seems to have its ID internally hence different hardware devices have different IDs which makes sense. 

The way to achieve that goal, however, might be manipulating `Omniscope::Id` or simpler abusing one of its `std::string`s.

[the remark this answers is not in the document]
AKMaily: I am not particularly concerned with how the integration within std::map is handled or the specifics of its functionality. On a broader level, the loaded data should depend on the filename rather than the device ID, as the device ID will likely be the same across most of the loaded files. I trust you to determine the best way to implement this.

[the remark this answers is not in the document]
R-Abbasi: Visiting Docs and being more familiar with the containers used in the project is worth doing since you'll gain a clearer overview of the features and also augment your skills that way.

The easiest way, as mentioned, is to embed the file name into one of the existing `std::string`s of `Omniscope::Id` e.g., `E4620C205B234D21 device1- ... .csv` as the `Id.serial`, I guess.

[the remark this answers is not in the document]
NP199: The Omniscopes have nothing to do with loading the files and i don`t understand why you want to combine loading data with the Hardware.
The software should handling the things needed for handling the files and if we just make a new map for handling loaded data it would make many things easier, since the device and file part are seperated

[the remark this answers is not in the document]
NP199: So if I understand the situation correctly, you want to have one container where all the data handling for displaying data is happening. To achieve this goal, we need to have the key-value pair from the std::map working. The key for the data stored is the Omniscope::Id right now, but the Omniscope::Id has nothing to do with the data loading part.

So if we want to load everything into a single map and handle it, we need to either modify the Omniscope::Id and add another identifier for the files being loaded, or we need to modify the handling of the files.

In my opinion, it is a bad idea to modify the Omniscope::Id to store information about the software handling its files.

Another approach could be to modify the std::map and change the key to a string, allowing us to use the Omniscope::Id::serial identifier for the connected devices and create a unique ID for the loaded files. The downside of this approach would be extensive refactoring and changes throughout the entire codebase. However, this might be necessary in the future when new devices are added.

Yet another approach could be to separate the data containers for the connected Omniscopes, the loaded data, and any future devices that may be added, and store them in different containers. The handling of the data containers would also need to be changed but could be unified in a generic approach where new containers can be added dynamically.

Are we on the same page with these ideas? @R-Abbasi  @AKMaily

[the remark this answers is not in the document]
R-Abbasi: Yes, but we need to display the data of the loaded files on the plot region for which the data should be inserted into `captureData` that still needs unique `Omniscope::Id`s.

src/style.cpp
```diff
@@ -499,7 +499,7 @@ void set_toolbar(const nlohmann::json &config, const nlohmann::json &language,
                                   image_height[PngRenderedCnt] * iconsSacle))) {
       for (const auto &[device, values] : captureData)
         if (!loadedFiles.contains(device))
-          liveDvcs.emplace(device, values); // extract live devices data
+          liveDvcs.emplace(device, values); // extract live devices (the little overhead)
```
R-Abbasi, line 502: The little overhead here would be less than 10 hits for `contains` and the same for `emplace` if 1000 devices/files are involved which is quite fast, of course there's some extra overhead for copying the _y_values_. There might be remedies to enhance this and reduce even that overall little overhead though.

src/saves_popup.cpp
```diff
@@ -10,9 +10,11 @@
 
 static void save(const Omniscope::Id &device,
                  const std::vector<std::pair<double, double>> &values,
-                 const fs::path &outFile, std::string allData, size_t &y_indx) {
-  allData += fmt::format(",{},{},{}\n",
-                         device.type, device.serial, device.sampleRate);
+                 const fs::path &outFile, std::string allData, size_t &y_indx,
+                 std::string filename) {
+  std::string serialFilename = device.serial + " " + filename;
+  allData += fmt::format(",{},{},{}\n", device.type, serialFilename,
```
AKMaily, line 16: IF the format of the header changes this should also be applied in the generateTrainingdata.hpp as this influences the data that is send to the API.

src/saves_popup.cpp
```diff
@@ -85,18 +87,20 @@ void saves_popup(nlohmann::json const &config, nlohmann::json const &language,
   ImGui::NewLine();
 
   if (ImGui::BeginCombo("##Combo", appLanguage[Key::Devices_Menu])) {
-    std::stringstream ss;
+    //std::stringstream ss;
     for (size_t i = 0; i < devicesSz; i++) {
-      ss << "Device " << i + 1;
-      ImGui::Checkbox(ss.str().c_str(), &(dvcCheckedArr[i].b));
-      ss.str(std::string());
+      //ss << "Device " << i + 1;
```
AKMaily, line 92: if this is not needed anymore please delete the comment

src/saves_popup.cpp
```diff
@@ -143,18 +147,17 @@ void saves_popup(nlohmann::json const &config, nlohmann::json const &language,
...
         selectedPathArr[i].clear();
         inptTxtFields[i].clear();
       }
-
       ImGui::CloseCurrentPopup();
     }
     ImGui::SameLine();
 
-    if (ImGui::Button("Close"))
+    if (ImGui::Button("OK"))
```
AKMaily, line 160: This should also be a Key::

AKMaily (commented):
First review without testing the application

src/saves_popup.cpp
R-Abbasi: What part needs changes exactly, please?

src/main.cpp
```diff
@@ -114,29 +110,35 @@ int main() {
...
-        loadedDvc = {};
-        loadedFileChkBx = false;
-      }
-    }
+    if (!loadedFiles.empty()) { // if devices were successfully loaded from file
+      static std::map<Omniscope::Id, BoolWrapper> loadedFilesChkBxs;
+      for (auto it = loadedFiles.begin(); it != loadedFiles.end();) {
+        ImGui::PushID(&it->first); // make unique IDs
+        if (ImGui::Checkbox("##", &loadedFilesChkBxs[it->first].b))
+          if (loadedFilesChkBxs[it->first].b) { // if checked
```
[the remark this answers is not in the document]
R-Abbasi: Have a look at [what](https://github.com/skunkforce/OmniView/blob/master/src/popups.hpp#L14) we've used for quite a while, please.

src/saves_popup.cpp
R-Abbasi: I left those keys to the one who made those changes to complete them himself and didn't touch that part much.

src/main.cpp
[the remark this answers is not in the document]
AKMaily: I see how defined it but "b" does not seem to be good variable name as it is not clear in the code what exactly "b" stands for.

src/saves_popup.cpp
AKMaily: In the generateTrainingdata.cpp the file is read but only the first line will be used as metadata. The way the data is currently saved not all information is in the first line. Therefore some of the metadata will be interpreted as doubles i guess.

src/saves_popup.cpp
AKMaily: As you have already changed the word, could you please make it a key to ensure the output in the software is correct?

src/saves_popup.cpp
R-Abbasi: Still don't see your point, sorry. 
`generateTrainingdata.cpp` only uses the first three elements of the first line (_Measurement_, _VIN_ and _Mileage_) and _y_values_; it doesn't touch the `serial` data member.

src/main.cpp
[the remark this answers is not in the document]
R-Abbasi: `if` needs a `bool` mandatorily.

src/saves_popup.cpp
AKMaily:
```

src/saves_popup.cpp
R-Abbasi: It doesn't seem to be needed as _OK_ is the same in both languages.

AKMaily: There will be more languages, so the code must be standardized accordingly.

src/saves_popup.cpp
R-Abbasi: [Have another look](https://github.com/skunkforce/OmniView/blob/feature-loading-old-data-second-iteration/src/saves_popup.cpp#L15), please. All metadata is written on the first line.

AKMaily (approved):

PR #168, opened 2024-06-28
Np/read calibration values v2

NP199:
In this pull request, the goal was to support two features. 
1. Receive Metadata from an OmniScope and save them for each device and the calculate the corresponding EGU (Engeneering Unit) to the graph to interpret the ADC values as a Voltage or whatever the EGu says.
2. With more then one EGU it would need more then one Y Axis to handle different y axis data. This shall be loaded automatically and when an device with no metadata is loaded it should be displayed aswell.


R-Abbasi:
- The above one represents the current behavior while the one below it representing your branch's. Is that the output you expected - as I talked about a couple of times - please?  



- It seems that lines below can be removed in `main.cpp`: 

 ```
  auto now = std::chrono::system_clock::now();
  std::time_t now_time_t = std::chrono::system_clock::to_time_t(now);
  std::tm now_tm = *std::gmtime(&now_time_t);
```

- It looks you used `|` instead of `,` in `ImPlot::SetupLegend` in `main.cpp`
- You also have unused variables `auxFlagsMeasuring` and `auxFlagsPaused` there.
- As well as, it seems you didn't apply the clang-formatter for all the files
- ...

**Review by R-Abbasi — COMMENTED** [review]
- The above one represents the current behavior while the one below it representing your branch's. Is that the output you expected - as I talked about a couple of times - please?  



- It doesn't seem necessary to `include` headers `algorithm` and `set`
- I think lines below can be removed in `main.cpp`: 

 ```
  auto now = std::chrono::system_clock::now();
  std::time_t now_time_t = std::chrono::system_clock::to_time_t(now);
  std::tm now_tm = *std::gmtime(&now_time_t);
```

- It looks you used `|` instead of `,` in `ImPlot::SetupLegend` in `main.cpp`
- You also have unused variables `auxFlagsMeasuring` and `auxFlagsPaused` there.
- As well as, it seems you didn't apply the clang-formatter for all the files
- ...

AKMaily:
Testing the Application: 

Case 1: OmnAIScope without calibration, signal generator frequency: 1 Hz

image

The time is set correctly, the Unit for the y-axis is ADC Counts as expected. 
I read that ADC Counts could only be positiv values, what is the reason that in this case the ADC Counts are also negative? 


Case 2: Calibrated Scope, signal generator frequenz: 1 Hz, Amplitude: 5 V

image

The Voltage is right, as well as the frequency

Case 3: Calibrated Scope and not calibrated scope are connected at the same time : 

image

As expected two different axis are shown. One for the ADC Counts and one for the voltage. The values are correct. 

Calibrated values are shown but not saved correctly: 

image
 
This should be fixed in another PR, as it is a seperate problem.

R-Abbasi:
My signal generator settings for the first screenshot was:
_Wave : **SINE**
FREQ: **20 KHZ**
AMPL: **2.5 V**
OFFS: **1.6 V**
DUTY: 50 % - no action
PHAS: **90.9 "**_
What are your full settings - like above - for your first two screenshots, please? (In order to reproduce those outputs)

AKMaily:
For the first screenshot they were: 

Wave: SINE 
Freq: 1 Hz 
Ampl: 5 V 
Offset : 0 V 
Duty: 50 % 
Phas: 0 

For the second they were : 

Wave: Rectangular 
Freq: 1 Hz 
Ampl: 5 V 
Offset: 0
Duty: 50 %
Phas: 0

R-Abbasi:
The branch doesn't compile on my system. Not sure how it passed the CI tests. Some of the errors:



None of the master or this branch version produced the first screenshot output using the settings above on my system. Not sure what version you've used for that.

AKMaily:
> The branch doesn't compile on my system. Not sure how it passed the CI tests. Some of the errors:
> 
> 
> 
> None of the master or this branch version produced the first screenshot output using the settings above on my system. Not sure what version you've used for that.

Which version of the omnai_communication repo did you use ?

R-Abbasi:
Compiled the project then switched to that branch and compiled it too. 
The errors are valid because I already spotted them when reading part of the changes made into that branch.

AKMaily:
> Compiled the project then switched to that branch and compiled it too. The errors are valid because I already spotted them when reading part of the changes made into that branch.

Which branch of the omnai_communication repo did you use as a submodule for this branch ?

AKMaily:
> Compiled the [OmniView](https://github.com/skunkforce/OmniView) repo for that branch. Don't see a specific way of compiling the branch mentioned somewhere.

You need to use the np/calibration branch of the communication repo to build the project. I think NP forgot to mention that in the description of the PR. It normally should switch when you are using: git submodule update --init --recursive while being in the PR branch

R-Abbasi:
Compiled the branch. The outputs using the signal generator settings above:
Wave: Sine


Wave: Rect

src/main.cpp
```diff
@@ -88,34 +91,47 @@ int main() {
...
-                                ImGuiCond_Always);
-      } else {
-        ImPlot::SetupAxes("x [Seconds]", "y [Volts]");
-        ImPlot::SetupAxesLimits(0, 10, -10, 200);
+    // Axes 1 to 3
+    // Check if time base for axes are same
+    // check if egu and timescale for plot are same
+    // error if third device is added
+    addPlots("Recording the data", flagPaused, [&flagPaused](double x_max, std::string yLabel, ImAxis_ axis, double yMin, double yMax) {
+        ImPlot::SetupLegend(ImPlotLocation_NorthEast |
```
R-Abbasi, line 104: `ImPlot::SetupLegend(ImPlotLocation_NorthEast, ImPlotLegendFlags_Outside);` was meant here likely.

src/main.cpp
```diff
@@ -88,34 +91,47 @@ int main() {
...
+                            ImPlotLegendFlags_Outside);
+        auto auxFlagsMeasuring =
```
R-Abbasi, line 106: As there's already a warning from CI which is ignored seemingly, `auxFlagsMeasuring` is initialized but not referenced.

src/main.cpp
```diff
@@ -88,34 +91,47 @@ int main() {
...
+            ImPlotAxisFlags_AutoFit | ImPlotAxisFlags_NoGridLines;
+        auto auxFlagsPaused = ImPlotAxisFlags_NoGridLines;
```
R-Abbasi, line 108: The same for `auxFlagsPaused`.

src/handler.cpp
```diff
@@ -1,133 +1,228 @@
...
 
-  if (ImPlot::BeginPlot(name, plotRegion, ImPlotFlags_NoFrame)) {
-    double x_min = std::numeric_limits<double>::max();
-    double x_max = std::numeric_limits<double>::min();
-    for (auto const &plot : plots)
-      if (!plot.second.empty()) {
-        x_min = std::min(x_min, plot.second.front().first);
-        x_max = std::max(x_max, plot.second.back().first);
-      }
+void addPlots(const char *name, bool const flagPaused,
```
R-Abbasi, line 53: `flagPaused` doesn't seem to be used here.

src/handler.cpp
```diff
@@ -1,133 +1,228 @@
...
+                ImPlot::PlotLine(
+                    fmt::format("{}-{}", plot.first.type, plot.first.serial)
+                        .c_str(),
+                    std::addressof(start->first), std::addressof(start->second),
+                    static_cast<std::size_t>(std::distance(start, end)) /
+                        stride,
+                    0, 0, 2 * sizeof(double) * stride);
+            }
+        };
+        for (int count = 0; auto const &plot : plotAxes) {
```
R-Abbasi, line 128: `count` looks like unused.

src/handler.cpp
```diff
@@ -1,133 +1,228 @@
...
+void addPlots(const char *name, bool const flagPaused,
+              std::function<void(double, std::string, ImAxis_, double, double)>
+                  axesSetup) {
+    static std::set<std::string> firstRun;
+    auto const plotRegion = ImGui::GetContentRegionAvail();
+    static int activeAxes{0};
+    //TODO search devices must work aswell
+    if (plotAxes.size() <= activeAxes) {
+        plotAxes = getDeviceInfos();
+        activeAxes = plotAxes.size();
```
R-Abbasi, line 62: Rather than 
```
 static size_t activeAxes;
 // TODO search devices must work aswell
 if (plotAxes.size() <= activeAxes) {
   plotAxes = getDeviceInfos();
   activeAxes = plotAxes.size();
 }
```
we could likely have 
```
 // TODO search devices must work as well
 if (plotAxes.empty())
   plotAxes = getDeviceInfos();
```

src/handler.cpp
```diff
@@ -1,133 +1,228 @@
...
-          auto p = std::upper_bound(start, plot.second.end(),
-                                    std::pair<double, double>{limits.X.Max, 0});
-          if (p != plot.second.end())
-            return p + 1;
-          return p;
-        }();
+        for (auto const &axes : plotAxes) {
+            // fmt::print("data size:{}, egu: {}\n", axes.data.second.size(),
+            //           axes.egu.first);
+            if (!axes.data.second.empty()) {
```
R-Abbasi, line 72: Rather odd to me, how `axes.data.second.empty()` could be empty while `plotAxes` has data in advance.

src/handler.cpp
```diff
@@ -1,133 +1,228 @@
...
-        auto const start = [&]() {
-          auto p = std::lower_bound(plot.second.begin(), plot.second.end(),
-                                    std::pair<double, double>{limits.X.Min, 0});
-          if (p != plot.second.begin())
-            return p - 1;
-          return p;
-        }();
+    }
+    if (ImPlot::BeginPlot(name, plotRegion, ImPlotFlags_NoFrame)) {
+        double x_min = std::numeric_limits<double>::max();
```
R-Abbasi, line 66: Is `x_min` going to always be `std::numeric_limits<double>::max();`, please? It doesn't seem to change.

src/handler.hpp
```diff
@@ -16,9 +28,12 @@ inline std::set<std::string> savedFileNames; // unique and ordered filenames
 inline std::optional<OmniscopeSampler> sampler{};
 inline std::map<Omniscope::Id, std::vector<std::pair<double, double>>>
     captureData;
-void addPlots(const char *, std::function<void(double)>);
+
+inline std::vector<AxisInfo> plotAxes;
```
R-Abbasi, line 32: `plotAxes` is used only in `handler.cpp`. It doesn't make much sense to have it in the header, I assume.

src/handler.cpp
```diff
@@ -1,133 +1,228 @@
...
 #include <implot.h>
 #include "popups.hpp"
 #include "get_from_github.hpp"
 #include "../imgui-stdlib/imgui_stdlib.h"
 
-void addPlots(const char *name, std::function<void(double)> axesSetup) {
-  static std::set<std::string> firstRun;
-  const auto &plots{captureData};
-  auto const plotRegion = ImGui::GetContentRegionAvail();
+std::vector<AxisInfo> getDeviceInfos() {
```
R-Abbasi, line 9: That must be `static`, I think.

src/handler.cpp
```diff
@@ -1,133 +1,228 @@
...
+            auto id = device.first->getId();
+            if (id.has_value()) {
+                Omniscope::Id deviceId = id.value();
+                std::string timebase{std::to_string(deviceId.sampleRate)};
+                if (captureData.find(deviceId) != captureData.end()) {
+                    auto eguIterator = std::ranges::find(
+                        assignedEgus, egu,
+                        &std::pair<std::string, ImAxis_>::first);
+                    if (eguIterator == assignedEgus.end()) {
+                        if (assignedEgus.size() <= 3) {
```
R-Abbasi, line 25: There's the comment "_The values assigned may change_" regarding `ImAxis_` in the Docs, therefore I suggest using also an assert like `static_assert((ImAxis_Y1 + 1) == ImAxis_Y2)` here.

src/handler.cpp
```diff
@@ -1,133 +1,228 @@
...
+                                "{}\nDevice id: {}",
+                                egu, id.value());
+                            break;
+                        }
+                    }
+                    AxisInfo axisInfo{
+                        std::make_pair(deviceId,
+                                       std::ref(captureData[deviceId])),
+                        *eguIterator, timebase};
+                    axisInfos.push_back(axisInfo);
```
R-Abbasi, line 43: `std::ref` here doesn't sound required. We likely could use a simpler version instead:
```
axisInfos.push_back({{deviceId, captureData[deviceId]},
             *eguIterator, timebase});
```

PR #175, opened 2024-09-02
Feature save calibrated values2

R-Abbasi:
- implemented saving different values based on the new range


AKMaily:
Tested the application: 
image

The values are now saved as correct values but they dont have the same amount of digits after the comma.

R-Abbasi:
They belong to useless trailing zeros in the decimal point section, I believe, E.g., **4.52100 -> 4.521**

PR #179, opened 2024-09-16
1st setp in refactoring

R-Abbasi:
This `pr` (second refactoring) is going to include:
- code cleanups
- code improvements
- documentation

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -12,21 +12,15 @@ int main() {
       load_json_file(load_json<std::string>(config, "languagepath") +
                      load_json<std::string>(config, "language") + ".json");
   // local variables
-  auto now = std::chrono::system_clock::now();
-  std::time_t now_time_t = std::chrono::system_clock::to_time_t(now);
-  std::tm now_tm = *std::gmtime(&now_time_t);
-  bool flagPaused{true};
-  bool Development{false}, flagInitState{true},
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -12,21 +12,15 @@ int main() {
...
   // local variables
-  auto now = std::chrono::system_clock::now();
-  std::time_t now_time_t = std::chrono::system_clock::to_time_t(now);
-  std::tm now_tm = *std::gmtime(&now_time_t);
-  bool flagPaused{true};
-  bool Development{false}, flagInitState{true},
-      open_generate_training_data{false}, open_settings{false};
+  bool flagPaused{true}, development{false}, open_generate_training_data{false},
+      open_settings{false};
+  std::once_flag flag;
```
AKMaily, line 17: Calling this flag rather configFlag or initFlag would be a better description as flag does not describe the use of the flag in the code.

R-Abbasi: It was used once that was why it sounds fine to me but `configFlag`/`initFlag` is fine too.

AKMaily: Please change the name to configFlag or initFlag, than we can merge the PR in the Devv1.0.1 Branch

AKMaily (approved):
Approved

